PR #10, opened 2016-07-13
Creates javascript.md

dotcode:


danyalaytekin:
ESLint is absorbing JSCS.

andrewmee:
Looks good to me.

@danyalaytekin I'd be inclined to wait until they have been absorbed until we change this, based on how they're saying they're progressing: http://eslint.org/blog/2016/04/welcoming-jscs-to-eslint  Unless someone fancies doing the gap-analysis work to see if our current rule-set is fully supported by ESLint

danyalaytekin:
@andrewmee Agreed, although JSCS proper is not seeing progress, and is described as deprecated by its team in GitHub.

https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.9a9f9tnkz

https://github.com/jscs-dev/node-jscs

(Declaration of interest: used ESLint on Oscar 😏)

dotcode:
Interesting! Given that we will be moving to ESLint, I think this is one of those occasions where we document the "slightly aspirational" rather than the "current state".

Given the information above, if someone were starting a project afresh, I'd want them to use ESLint.

Saying projects _should_ use ESLint is OK, I think (I'd think differently if it were a "must").

PR #12, opened 2016-07-14
Removes "graceful degradation is fault-tolerant"

dotcode:
Graceful degradation is *broken* by design. Progressive enhancement starts from a position of "your site works on every single user agent". Graceful degradation starts from a position of "your site works on a handful of the tens of thousands of user agents that exist" … and there is no possibility of you ever catering for more than a handful of them.

### DIFF
--- a/practices/progressive-enhancement.md
+++ b/practices/progressive-enhancement.md
@@ -10,7 +10,7 @@ There is a subtle but important difference in this approach vs the approach of "
 
 There are several key benefits to using progressive enhancement;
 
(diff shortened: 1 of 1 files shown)


andrewmee:
I was attempting to be a bit more diplomatic towards the graceful degradation crowd, but happy to take a stronger stand against it 👍

dotcode:
Cool, ta. I don't see the two as being in competition, truth be told. Graceful degradation does not achieve it's aims, progressive enhancement does.

PR #15, opened 2016-07-14
adds some @jakearchibald

dotcode:
… because everything needs more @jakearchibald


dotcode:
( … just added @jakearchibald in the hope he can provide the :+1:)

benjclark:
Was it your intention for them to list underneath each other? At the mo the references are on the same line

dotcode:
Doh! Should have been in a list as well - thanks for spotting :)

benjclark:
👍

PR #16, opened 2016-07-15
javascript - adds "optimise for reading"

dotcode:


andrewmee:
👍 Hard one to stay vigilant about, but a good one to have

PR #20, opened 2016-07-21
Add outline of javascript directory structure

andrewmee:
Putting this forward as a proposal - please do feel free to offer alternative names, structures, definitions, etc.


dotcode:
LGTM

benjclark:
What are the SN.Component and SN.util objects? Does this refer to the global objects within a project that are responsible for those functions?

andrewmee:
@benjclark Just neat places to namespace our JS into which reflect what kind of JS it is (DOM-binding or non, etc.).  The idea being when you see a call for `SN.util.X` you've got a good idea about what kind of script that is referencing (and where to find it in the `resources/js/*` directory structure).

benjclark:
Thanks 👍

PR #21, opened 2016-07-21
First pass at OOCSS section

sky-jack:


practices/oocss.md
```diff
@@ -1,12 +1,39 @@
 # Object Oriented CSS
 
+
+
 ## What is OOCSS
 
+OOCSS or 'Object Orientated CSS' is an approach to writing css that aims to encourage reusable, maintainable and predictable css. It has spawned a number of different approaches to standardise the concept most of which adhere to the core principles of OOCSS outline below.
+
+A CSS 'object' is a repeating visual pattern, that can be abstracted into an independent snippet of HTML, CSS, and possibly JavaScript. That object can then be reused throughout a project. The aim is to break down the css declaration into the smallest blocks possible in order to maximise their potential for reuse.
```
dotcode: would like to change to "smallest blocks necessary to enable reuse" rather than "smallest blocks possible in order to maximise their potential for reuse". The smallest blocks _possible_ is the aim of atomic CSS, not OOCSS as a whole.

practices/oocss.md
```diff
@@ -1,12 +1,39 @@
...
+
+An important component of OOCSS is the process of defining repeating visual features as separate modifying 'skins' that you can apply as distinctive classes to allow visual variety regardless of the structure of the HTML. In order to do this css declarations should not rely upon any assumptions about the semantics of the HTML. Instead classes are used to define 'objects' and the components therein. The layout and display of particular object or components can then be modified by additional classes or 'skins'.
```
dotcode: "An important component of OOCSS is the process of defining repeating visual features as separate modifying 'skins' that you can apply as distinctive classes to allow visual variety regardless of the structure of the HTML." < is it possible to somehow break this sentence down?

practices/oocss.md
```diff
@@ -1,12 +1,39 @@
...
+
+
+The two guiding principles of OOCSS could be defined as follows:
+
+- Separation of structure from skin: Keep the visuality separate, so you can reuse the visual classes
```
dotcode: <looks up in dictionary/> Huh. I didn't even realise visuality was a word! Can we reword this a little so it's more something along the lines of "keep the appearance separate, so you can reuse those classes on different elements"?

practices/oocss.md
```diff
@@ -1,12 +1,39 @@
...
+
+- Separation  of container and content: Avoid the use of location-dependent styles and always aim to identify containers and components that can be reused.
+
+
+In this regard it privileges lack of repetition in the stylesheet over increases to the size of the HTML through the addition of extra classes.
+The hope is that in many cases the reduced css file size will out-weight the potential bloating of the HTML.
```
dotcode: Might be worth being a little stronger here - it's more than a hope.

In fact I'd recommend deleting this sentence. If you look at our html, you can't characterise it as "bloated". The only circumstance where I can see this actually being worthy of mention is if you had an incredibly high traffic site with multiple different page views where saving a few bytes on every page (through being able to cache all the css, and having smaller page weights due to fewer classes) could outweigh the maintainability benefits of OOCSS. So … like the traffic of Facebook, the variability of pages that we have on our site, and not using any form of edge caching.

"Reduced CSS size vs HTML bloating" isn't a thing :)

dotcode:
Jack this is amazing. Thanks so much for kicking this off.
💯

hollsk:
Some concrete examples would be helpful - maybe illustrate with an actual component that we've already put to use?

dotcode:
> Some concrete examples would be helpful

dotcode:
> Some concrete examples would be helpful
In the interest of getting this merged ASAP, shall we open an issue and deal with in a new PR?

sky-jack:
Thanks all for the feedback. I've just pushed an updated based on the suggestions.

danyalaytekin:
Nice work @sky-jack.  We just spoke and I realise now that (a) I've been conflating Atomic CSS and OOCSS a bit in my mind, and (b) I was thrown by the use of the word 'semantic' above, where I understand that it now belongs to the wider phrase 'semantics of the HTML' (rather than being a restriction on the use of terms describing page features in CSS classnames).  With @dotcode's clarification of the other sentence to "smallest blocks necessary to enable reuse" rather than "smallest _possible_", a FED reading this gains some discretion and could specialise to either BEM or Atomic.

sky-jack:
Thanks for the feedback Danyal, yeah I agree that the distinction between 'smallest blocks necessary to enable" and "smallest possible" makes all the difference and is a vital distinction.

sky-jack:
Perhaps we should expand the common concerns section to include something about the problems with reconciling "atomic css" and BEMM? Because these are concerns that may well become more important for us.

danyalaytekin:
@sky-jack Seems like this PR is good to go tbh? That work might deserve its own.

danyalaytekin:
Also need to think about whether BEM, Atomic and their descendents-presumably-hurtling-towards-us will belong in this document, or in their own.  And there are other CSS-related concerns such as editors, linters, browser plugins, build tools and so on which might not fit into `oocss.md`.  One option is having a `css` folder with those files in, or just a big `css.md`.  Again though I think this is PR is a great first attempt and can be merged.

dotcode:
Concur. Merge away.

PR #23, opened 2016-07-22
adds jump links

dotcode:


andrewmee:
I was just thinking we could use those... 👍

PR #26, opened 2016-07-22
adds "why we care"

dotcode:
DON'T MERGE! Still more work to do.


dotcode:
urgh, what am I doing?! Friday evening. Probably time to step away from the computer.

PR #27, opened 2016-07-22
adds performant code

dotcode:


nickcall:
Should there be a comment about commenting the code ?

dotcode:
Ooh good point - I think we do need that _somewhere_, but probably not in this section. Can you open an issue to get that done, and link to it here?

nickcall:
Will do that, so I'll sign off this PR without then

👍

PR #29, opened 2016-07-25
adds "indentation"

dotcode:


andrewmee:
Before the holy war kicks off, I'm going to get in early and say I am okay with this... I've used soft tabs before and hard tabs before, and honestly it doesn't bother me which - provided we use them consistently.

So it's a "LGTM 👍 " from me, as would the same addition using soft tabs if that's what others prefer.

hollsk:
What Andrew said, anything is fine so long as it's consistent 👍

PR #31, opened 2016-07-25
adds copyright notice

dotcode:
… to comply with the licensing requirements detailed at https://www.w3.org/WAI/about/usingWAImaterial#cc


andrewmee:
Looks good

PR #33, opened 2016-07-25
adds "why we care about accessibility"

dotcode:


andrewmee:
Looks good, but I wonder if we could go further...  Much of our content is educational in nature; if we know that people can encounter difficulties accessing our content (which we do), and we know that we can do simple things to facilitate that access (which we do), then **not** making something accessible is knowingly contributing to the oppression of people with disabilities.

dotcode:
Ooh, nice. Happy to add something like that, although unsure of the tie to educational content?

andrewmee:
We probably don't need the tie to educational content, it just strikes me that hampering someone's access to any content is bad, but when that content is educational it makes it way, way worse (hampering someone's personal development, job prospects, etc.)

dotcode:
OK, cool. I'm happy without the qualification. Will add.

andrewmee:
👍

PR #34, opened 2016-07-25
Tidy up semver 'find out more' links

andrewmee:


practices/semver.md
```diff
@@ -31,13 +31,6 @@ When a piece of software under heavy development many major/minor releases may h
...
-How we version and release our open-source software: https://github.com/springernature/shunter/blob/master/docs/developer-guide.md#versioning
-More reading: https://abdulapopoola.com/2015/10/26/what-is-semver/
-
-
-
-
-
-
-
+- Canonical source of info: http://semver.org/
```
dotcode: rather than having some text, with a url following, can we instead have sensible linked text? Is more accessible.

practices/semver.md
```diff
@@ -31,13 +31,6 @@ When a piece of software under heavy development many major/minor releases may h
...
+- [How we version and release our open-source software](https://github.com/springernature/shunter/blob/master/docs/developer-guide.md#versioning)
+- More reading: https://abdulapopoola.com/2015/10/26/what-is-semver/
```
dotcode: rather than having some text, with a url following, can we instead have sensible linked text? Is more accessible.

sky-jack:
fine title links 🎱

PR #37, opened 2016-07-26
add variable styles

dotcode:


howlingmad:
Looks good 👍

danyalaytekin:
Looks good.  Something to consider for another section/document is ES6.  On Oscar when using ES6 we eradicated `var` in favour of `let`, as well as using `const` wherever possible.

dotcode:
Good point @danyalaytekin - have created #48

PR #38, opened 2016-07-26
qualification of graceful degradation as outdated

dotcode:
because … you know … it was outdated as soon as progressive enhancement became a thing.


howlingmad:
👍  makes sense

PR #39, opened 2016-07-26
add documentation for BEM best practices and methodology

howlingmad:
- Adding documentation on how to write BEM.
- This is mostly taken from my project so would be good to get agreement from others that use BEM that they are happy with this.


davidpauljunior:
That matches our approach too 👍

danyalaytekin:
Good stuff

PR #40, opened 2016-07-26
Remove links to removed sections, add section on preprocessors

howlingmad:
- Top of the document linked to sections that have been moved
- Added a section on preprocessors


dotcode:
Could we see some "we do this/we don't do this" examples?

howlingmad:
how would that work in this context?

dotcode:
Looks good.

PR #41, opened 2016-07-26
update the naming style to match mosaic

howlingmad:
Update the way we talk about naming conventions to match the style used in mosaic-css.md


davidpauljunior:
👍

PR #42, opened 2016-07-26
add "we don't/do this"

dotcode:


howlingmad:
nice 👍  will need to update the mosaic-css naming conventions as well

languages/bem-css.md
```diff
@@ -18,13 +18,35 @@ BEM naming follows this pattern:
...
+
+We _don't_ do this:
+```css
+.feature
+.content
+.btn
+```
+
+We do do this:
+```css
```
howlingmad: "do do"

a51

PR #46, opened 2016-07-27
style rules for css

howlingmad:
I wrote these rules for SpringerLink, but figured they were pretty universal, so have ported them over. Shout at me if you don't agree 😁


davidpauljunior:
The shorthand is an interesting point. I wouldn't use shorthand for something like `font`, but I would use it for `border`.
```border-width: 1px; border-style: solid; border-color: green;``` seems cumbersome for some reason.  Then I guess that is inconsistent usage 😕

howlingmad:
yeah, maybe i need to re-word it. To me it is fine to use shorthand in this case as you are setting all the values. The issue is setting say _just_ colour using `background`, as it _might_ unintentionally overwrite another rule.

PR #48, opened 2016-07-28
add ES2015 variables

dotcode:


hollsk:
Looks great. Should we also specify strict mode? I don't know what the Shunter convention is, but it's generally recommended elsewhere.

danyalaytekin:
:+1:

rowanmanning:
Something I'm doing now, which might be worth adding here if you agree, is using `const` everywhere unless a variable _explicitly_ needs to be modified. It's great for the default state of a variable to be immutable, and I've found that I actually have very little need to use `let`.

There's some good background here: https://ada.is/blog/2015/07/13/immutable/

dotcode:
Thanks @rowanmanning - better now?

PR #57, opened 2016-08-02
add ES2015 classes

dotcode:


joseluisbolos:
Why not:

```
class Apple extends Fruit {
    constructor() {
        this.color = 'red';
    }

    peel() {
        // ...
    }
}
```
???

andrewmee:
We don't mandate empty lines at the top and bottom of functions, so I'd tend to agree with Jose; I don't think we should care if there are empty lines within that class declaration.

PR #59, opened 2016-08-02
add the standards we comply with

dotcode:


danyalaytekin:
There's also a part of line 20 which mentioned this - it can probably be removed.

PR #63, opened 2016-08-09
create accessibility checklist

dotcode:


hollsk:
Looks perfect. One link (to the effective colour contrast) points to an internal resource and should go somewhere open, though. 

I'm hemming and hawing a bit as to whether we should recreate that internal resource here or not - we can't link to the source text anymore because it's linkrotted. On the one hand the information is really useful, but on the other hand it'd open the door to a massive accessibility guide. That in itself isn't a bad thing by any means, but is it in scope for _this particular project?_

dotcode:
Ooh, good catch.

<looks in vain for the "thinking-face" emoji />

Hmm. Yeah … I want to avoid creating an accessibility guide. The reason I created that page on the wiki was because it was just _so_ useful. Illustrates the issue really well, seems right to ...

Right, so I've argued myself into recreating that page here. It's right that the information exists on the web, someone else might find it useful.

PR #69, opened 2016-08-18
fix incorrect link

dotcode:
addresses https://github.com/springernature/frontend/pull/68#discussion_r75278145


andrewmee:
I half expected to be rickrolled then, but it checks out...

dotcode:
I shall rue this missed opportunity until the day I die.

PR #72, opened 2016-08-22
add javascript "Operators" section

dotcode:


andrewmee:
👍

languages/javascript.md
```diff
@@ -358,6 +358,31 @@ class Apple extends Fruit{
 }
 ```
 
+### Operators
+
+All operators must have padding whitespace on either side of them. The only exceptions are increment (`++`), decrement (`--`), unary negation (`-foo`), and unary plus (`+foo`).
+
+We disallow the use of normal equal and not equal, requiring you to use the strict equivalents.
```
danyalaytekin: This could be written in the tighter style of the previous paragraph e.g. "We forbid the lenient equality operators `==` and `!=`."

Issue #75, opened 2016-08-22
add "we do/don't do" links to "blocks" section


dotcode:
The [blocks section in the JavaScript guidelines](https://github.com/springernature/frontend/blob/master/languages/javascript.md#blocks) is missing the helpful "we do/don't do" code examples of other sections. Would be good to add.

PR #79, opened 2016-08-24
add javascript "Error handling" section

dotcode:
Create error handling subsection under "Blocks", re-title previous "Error handling" section as "Strict mode".

Addresses #75

### DIFF
--- a/languages/javascript.md
+++ b/languages/javascript.md
@@ -11,7 +11,6 @@ Projects should use both [JSHint] and [ESLint] to enforce these rules.
   - [Write modules over monoliths](#modules-over-monoliths)
   - [KISS](#keep-it-simple)
   - [Performant Code](#performant-code)
-  - [Error handling](#error-handling)
 - [Code Style](#code-style)
   - [Indentation](#indentation)
   - [White Space](#white-space)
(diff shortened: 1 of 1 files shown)


hollsk:
"You should assume that your code will fail, and take steps to handle those failures" would make more sense at the start of the "error handling" subsection, i reckon - it's an introductory phrase, really. Otherwise 👍

dotcode:
… although … do you think it's a better introductory phrase to the strict mode stuff than the try/catch stuff?

hollsk:
Not strictly (haha see what i did there? geddit?!). Adding a try/catch is literally, by definition, assuming that your code will fail, and writing something to handle that assumed failure. 

IMO "error handling" and "strict mode" are conceptually linked, as the purpose of strict mode is to enable better/saner error handling.

dotcode:
Yeah, I agree with all that. I suppose … I think it's a better introductory phrase to the strict mode stuff than it is to the try/catch stuff, because the try/catch stuff is purely about formatting, whereas the strict mode is more about methodology.

dotcode:
those two sections - as they currently exist - obviously could be written in a better manner (as evidenced by this discussion :)

I've had an idea (my only one of the week, so be gentle). How about moving the try/catch stuff out of "Blocks" and sticking it under "Error handling"? Would that work?

hollsk:
¯\_(ツ)_/¯

The confusion is down to me not realising the strict mode section existed in the Nature Playbook, so I've reinvented the wheel. The difference between my commit and the previous work is just that I've interpreted the concept of strict mode as being a subset of "error handling", where the other author treated them as separate concepts. 

You could remove the entire controversy by deleting my intro line entirely, which would bring into line with the Nature Playbook :p

dotcode:
OK. I like your introduction. Works well as an introduction to the "strict mode" section, so going to just merge this. We can always revisit if we come up with anything better.

PR #80, opened 2016-09-05
Add inclusive language guidelines

joseluisbolos:
This PR adds new language guidelines for inclusive, non-discriminatory language.

I would really appreciate if @hollsk and @typothecary could have a look at it. ;)

PR #82, opened 2016-09-05
Replace backticks with italics

joseluisbolos:
Also removes the NYT reference and that was in relationship to the _Mx_ honorific removed in #80


danyalaytekin:
:+1:

PR #84, opened 2016-09-09
mv languages technologies

jpw:


jpw:
Discussion as per https://github.com/springernature/frontend/issues/83

dotcode:
👍

PR #87, opened 2016-09-15
Markup

jpw:


technologies/markup.md
```diff
@@ -0,0 +1,48 @@
+# Markup
+
+## Background
+
+HTML up to 3.2 was written with SGML syntax rules.  HTML 4 could be written with either SGML syntax (HTML 4.01) or XML (XHTML 1.0).  [HTML 5 breaks with SGML and has two serializations](https://www.w3.org/blog/2008/01/html5-is-html-and-xml/), a new one called "html" (which looks like SGML but is not an application of SGML) and again XML (XHTML).
```
umaar: This is really great as a whole. Thanks! I am wondering if we can make it more concise. We want the playbook as a whole to be a compelling read, therefore I'm wondering if this background is needed? 

If you feel it is, what are you thoughts on adding it to the bottom as a 'Further Reading' section?

technologies/markup.md
```diff
@@ -0,0 +1,48 @@
...
+**We do not omit optional end tags.**
+
+While the HTML 4 & 5 specs allow some end tags to be omitted for _some_ non-[void elements](https://www.w3.org/TR/html5/syntax.html#void-elements), it's unreasonable to expect authors to remember which elements this applies to.  Additionally this kind of inconsistency can play havok with text editors, any indentation policy, and impairs our ability to spot bugs that would otherwise offend our pattern-recognition powers.
+
+
+## Void Elements -- to explicitly or implicitly close?
+
+**We implicitly close void elements.**
+
+Explicitly closing void elements (e.g. `<br />`) is not invalid per se in HTML 5.  However, explicitly closing void elements is an artifact of XML serialization, not html, so may introduce cognitive dissonance in many authors given that we are authoring html.  Another source of dissonance is that explicitly closing void elements was invalid in older versions of HTML.
```
umaar: Change `html` to `HTML`

technologies/markup.md
```diff
@@ -0,0 +1,48 @@
+# Markup
```
umaar: This whole doc is an interesting read! Please could you add a tl;dr at the top and summarise the primary piece of information a reader should take away from this file?

technologies/markup.md
jpw: I think it is needed because subsequent passages refer to the serializations, and it's stuff a lot of ppl think they know, but don't really :)

jpw: i do agree the whole page is a bit long though

technologies/markup.md
jpw: It's "html" as in the serialization, but yes, it looks wrong -- I  should be more specific there

danyalaytekin:
Great start, and I'm happy to settle on this convention.  Some thoughts though:

a) the document is emphatically - almost evangelically - in favour of non-self-closing tags.  I don't remember this being the consensus during our group discussion, which was a compromise decision by majority, rather than a unanimous decision, with the majority of our group abstaining from the vote.

b) it doesn't acknowledge the arguments in favour of self-closing tags (e.g. some of us find them easier to read).

c) I agree with @umaar that the conclusion should appear separately and briefly at the top, with the rest of the background in a separate section below, or removed.

dotcode:
Concur with all this. Requested changes sound sensible. Excellent work.

jpw:
- a summary of the document added, and an index
- as I mention, the Background bit is there to explain the html-not-SGML serialization I subsequently refer to in the doc, so to me it makes sense to keep it there at the top.
- "a) the document is emphatically - almost evangelically - in favour of non-self-closing tags" I must admit I don't see it as that emphatic, but please feel free to suggest rewording.
- "b) it doesn't acknowledge the arguments in favour of self-closing tags" I do this in the last sentence ;) again, feel free to soften/amend/add
- "c) I agree with @umaar that the conclusion should appear separately and briefly at the top" by that do we mean a document summary? or a summary of all the points?

jpw:
Oh and re "html casing correction" I have rewritten that bit so it looks less like a mistake :)

umaar:
Is looking really good! 

> c) I agree with @umaar that the conclusion should appear separately and briefly at the top" by that do we mean a document summary? or a summary of all the points?

What I mean is, there's quite a bit of reading to do in that document, however not everyone may wish to do it! If you could communicate to the reader at least one (or two) things they should take away (e.g. an action point), what would it be? For me, the key points are:
* We do not omit optional end tags.
* We implicitly close void elements.

The reason I would prioritise those over:
* We author HTML 5.
* We do what is reasonable to ensure our documents validate .

Is because the first two are slightly more opinionated, whereas those two bullet points above, I don't think anyone would disagree with (and I think most web folk strive for that anyway?)

So a small section at the top titled `tl;dr;` or `Summary of action points` with those original two points (if you agree) might be fine. 

I would also add a small inline code snippet next the bullet point(s) at the top to explain exactly what we mean.

jpw:
I see, good point thanks @umaar ! As for web folks being in agreement on XHTML etc, one of the reasons I got carried away in this doc going on about XHTML is that FEDs are not the only audience for this -- it's also for the benefit of engineers who feel that somehow XHTML is intrinsically preferable ;)

umaar:
This is great! Thanks for making all those changes :

danyalaytekin:
I have flip-flopped, and am now in favour of this, realising that it is more internally consistent than the alternative.  The thing which convinced me was the effect when applied to a non-void element: a conversion to an opening tag, which is actually dangerous.

PR #93, opened 2016-09-23
Add colon to css pseudo-element selector

### LINKED ISSUE #92: : VS ::
Hi,

I see in your CSS documentation that you use ":" for your pseudo-elements (before, after) as for your pseudo selectors (hover, focus).

CSS 3 specifications say that pseudo-elements should be differenciated with two colons "::"
See this [article](https://css-tricks.com/almanac/selectors/a/after-and-before/#article-header-id-0) from CSS Tricks.

Single colons are often used to support IE8 though. But the correct syntax is "::" for pseudo-elements.

dotcode:
Fixes #92

### DIFF
--- a/technologies/css.md
+++ b/technologies/css.md
@@ -127,7 +127,7 @@ a {
 }
 
 span {
-    &:before {
+    &::before {
         content: '\2022';
     }
 }


hollsk:
:: 👍 excellent update

Should we open a separate issue to add a bit about ::ing, now that we're able to make it formal in our apps?

dotcode:
Not even sure we need that tbh - we're just following the spec.

PR #94, opened 2016-09-29
Restructure markup file and document heading level strategy

andrewmee:
Attempting a structure and language use that's closer to what we use elsewhere in this playbook

### DIFF
--- a/technologies/markup.md
+++ b/technologies/markup.md
@@ -11,7 +11,35 @@ This document outlines the way we write markup and why.
 
 ## Summary Of Action Points
 
-- We do not omit optional end tags [(discussion)](#to-omit-optional-end-tags).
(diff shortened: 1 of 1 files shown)


technologies/markup.md
```diff
@@ -39,47 +69,39 @@ We *don't* do this;
...
+		<h3>Article subtitle</h3>
+	</article>
 
-## Void Elements -- to explicitly or implicitly close?
+	<footer>
+		<h2>Footer title</h2>
+	</footer>
 
-**We implicitly close void elements.**
+We *don't* do this;
```
danyalaytekin: Should these be `:`?

andrewmee: Yep!  Good catch

PR #101, opened 2016-10-10
Move all the `*css` files to `technologies`

joseluisbolos:
(Except `CSS` itself, which belongs to `languages`)


joseluisbolos:
Good catch. That's fixed now. 👍

andrewmee:
Oops, no you've linked to technologies rather than practices

joseluisbolos:
Good catch. That's fixed now. (2/2) 👍

Issue #32, opened 2016-07-25
Contribution guidelines


dotcode:
Detail how we contribute to the playbook. PRs, tone of voice, etc.

See "[Setting guidelines for repository contributors](https://help.github.com/articles/setting-guidelines-for-repository-contributors/)".

hollsk:
Now completed. There's a separate code of conduct issue ( #110 ) so I'm gonna close this one 👍

PR #111, opened 2016-10-27
Add a contributor guide

hollsk:
Resolves more of #32 

I think the only thing missing is tone-of-voice guidelines, which should probably go in the language section.


moddular:
👍 Looks good to me

PR #113, opened 2016-11-21
Minor JS updates, also added a link for ‘when to optimise’.

umaar:


technologies/javascript.md
```diff
@@ -545,11 +548,11 @@ In `for` loops, there must be space after the semi-colons.
 We do this:
 
 ```js
-for (var i = 0; i < 10; i += 1) {
+for (let i = 0; i < 10; i += 1) {
     // ...
 }
 
-for (var prop in obj) {
+for (let prop in obj) {
```
danyalaytekin: `const`?

umaar: Done

PR #124, opened 2017-01-31
Add versioning section to the git guide

joseluisbolos:
Ripped off from the Shunter docs, except for the new recommendation to
not use git push --tags to push all the tags on release.

### DIFF
--- a/practices/git.md
+++ b/practices/git.md
@@ -17,6 +17,7 @@ We do this:
 frontend-project-name
 ```
 
+
 # Commit Messages
 
 Your commit messages are your personal legacy.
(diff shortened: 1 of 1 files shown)


andrewmee:
I don't think that this really fits in our git practices document.  It only related to our open-source repositories, but that's a minority of the work we do.  It's also more instructional rather than dealing with best-practice as it's currently drafted.

If we want to add this info into the playbook I'd think it should go in a different section and/or would need to be rewritten.

joseluisbolos:
We definitely need this information somewhere - And I think that Shunter is the wrong place to have it. We also have it replicated in other repos (which is a maintenance nightmare) and it needs to be updated (as the recommendation to use git push --tags is wrong).

My intention was to add it here and then remove all the duplicates everywhere and replace them with a link to the playbook, so we only have one source of truth.

If people don't think this is an appropriate place to store this, can anyone suggests where would be an appropriate one?

umaar:
> This repo contains the Springer Nature front-end playbook. It details how we run software development and how we make web and mobile products together. 

I don't think having instructional material in the playbook is wrong. 

@andrewmee what do you think if we add a little "Please Note:" section at the top to explain to other folks that this is quite specific to us, but they're welcome to learn from and adapt to our Git versioning approach?

umaar:
Also just reading the git practices, it's reasonable stuff.

- Use PRs
- Delete remotes afterwards
- Use `semver`
- etc.

But it's fair to go into a different section, maybe just change the heading from `Versioning` to `Versioning of open source projects`?

andrewmee:
@umaar yeah some rewording to make it clear this is related to Open Source projects only.

It stikes me that we could use a general practice document specifically for Open Source project management, which deals with things like how we document and govern OS projects.

joseluisbolos:
Amazing that none of us has realised that this has been in the playbook since October :man_facepalming: 

https://github.com/springernature/frontend-playbook/blob/master/practices/open-source-support.md

PR #126, opened 2017-01-31
Replace train-case with lowercase-hyphenated case

### LINKED ISSUE #125: Train-Case
We seem to be using the expression `train-case` in [three different places in our playbook](https://github.com/springernature/frontend-playbook/search?utf8=%E2%9C%93&q=train-case), with the meaning _"all lowercase, words separated by a hyphen"_.

Wikipedia seems to differ in the meaning of [train-case](https://en.wikipedia.org/wiki/Letter_case):

> If every word is capitalised, the style is known as Train-Case.

Proposing to change the name to one that is understandable to everyone, specially for non-native English speakers, without having to google it first. Any suggestions?

joseluisbolos:
Fixes #125

### DIFF
--- a/practices/bem-css.md
+++ b/practices/bem-css.md
@@ -18,7 +18,7 @@ BEM naming follows this pattern:
 
 ## Naming Style
 
-All class names should be train-case - all lowercase, words separated by a hyphen. 
+All class names should be `lowercase-hyphenated` - all lowercase, words separated by a hyphen. 
 
 We _don't_ do this:
 ```css
--- a/practices/bem-css.md
+++ b/practices/bem-css.md
@@ -18,7 +18,7 @@ BEM naming follows this pattern:
 
 ## Naming Style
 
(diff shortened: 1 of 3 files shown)


nickcall:
LGTM

davidpauljunior:
Looks good, just wonder if it should be taken out of code formatting, as it's not actually a code snippet, it's a name.

`lowercase-hyphenated` or "lowercase-hyphenated"

joseluisbolos:
At risk of bike-shedding even more, I marked it up with a monospace font as to indicate that it's a... well... the name of a thing that we've just made up. But you're right, according to [O'Reilly's Typography and Font Conventions](http://chimera.labs.oreilly.com/books/1230000000969/ch03.html) it should probably be _Body font italic_ and not monospaced. The [Wikipedia page](https://en.wikipedia.org/wiki/Letter_case) also uses the same convention.

Is that acceptable?

PR #134, opened 2017-02-09
Revert recommendation to use `--follow-tags`

joseluisbolos:
This is because `--follow-tags` doesn't work with lightweight tags,
which are the ones that we use in our open source projects. Instead,
they only work with annotated tags.

Yes, there are two kinds of tags in git. The more you know... 💫


moddular:
Looks good to me 👍

Issue #66, opened 2016-08-15
Defining dependency versions


andrewmee:
It's probably a fairly minor point but I think it might be worth having a standard practice we use when defining dependency versions (for NPM-managed packages for example).

At the moment there are a range of ways that we specify versions - some are locked down to a particular release e.g. `1.2.3`, others to ranges of releases.  It could be useful to offer guidance on when it's preferably to use a range, and when it's better to lock to a specific version.

There are also several ways to define a range, which can cause confusion.  See https://github.com/npm/node-semver#advanced-range-syntax for a whole heap of examples.  I think we should aim to use the same range methods across the board (when they're equivalent and interchangeable).

andrewmee:
FWIW I favour being as permissive with range as possible, so if it's a dependency that uses semver correctly it should be safe to allow patch release updates automatically.  And I prefer tilde range specifications e.g. `~1.2`

hollsk:
Jose, is this issue ok to close?

joseluisbolos:
Yup, this was added to the Playbook in #139

PR #139, opened 2017-02-28
Add dependency version guidelines

joseluisbolos:
Addresses #66

### DIFF
--- a/practices/dependency-management.md
+++ b/practices/dependency-management.md
@@ -1,11 +1,11 @@
 # Dependency management
 
-The following guide describes how we manage package dependencies in our Node.js-based projects.
+The following guide describes how we manage package dependencies in [Springer Nature Node.js-based projects](https://github.com/springernature?utf8=%E2%9C%93&q=&type=public&language=javascript).
 
 
 ## Specifying versions of dependencies
 
-As a general rule, we tend to follow a conservative approach when specifying versions for the dependencies of our packages. This helps us ensure consistency between different environments, as we can't know for sure in what environment the app is going to be running, and prevent breakage caused by major or minor updates in dependencies.
+As a general rule, we tend to follow a conservative approach when specifying versions for the dependencies of our packages. We only allow PATCH version ranges for [run-time dependencies](#run-time-dependencies) and MINOR + PATCH version ranges for [dev dependencies](#development-dependencies). This helps us ensure consistency between different environments, as we can't know for sure in what environment the app is going to be running, and prevent breakage caused by major or minor updates in dependencies.
 
 We find that the this approach strikes a good balance between the potential breakage caused by less restrictive versions and the maintenance needed when restricting the dependencies to a specific version through the use of `npm shrinkwrap` or similar tools.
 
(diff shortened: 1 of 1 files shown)


alexkilgour:
Where you say we should be using tilde or carat, it could be useful to specify WHY that choice is made - describe what effect this method will have on what is downloaded

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
...
+We use some [dependency management tools](#dependency-management-tools) to help us keep our apps up to date when new versions of their dependencies are released.
+
+
+### Run-time dependencies
+
+[Run-time dependencies](https://docs.npmjs.com/files/package.json#dependencies) are modules that are required for your application to run, independently of the environment or mode (e.g. development vs production) used.
+
+These are defined in the `dependencies` section of the `package.json` file.
+
+Version numbers for run-time dependencies are defined using a tilde `~` plus a full version number in the MAJOR.MINOR.PATCH form. If the PATCH version is 0, it can be omitted.
```
dotcode: I suggest removing this: "If the PATCH version is 0, it can be omitted."

Should avoid unnecessary exceptions, as it makes for simpler documentation, and in this case will result in less deviation from the norm in the code.

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
...
+
+Specifying the versions in this way ensures that:
+* We don't expose the users of our modules to an unnecessary risk if one of our dependencies releases a minor update that breaks our app.
+* Dependencies in early lifecycle projects (v0.MINOR.PATCH) are traditionally considered of beta quality, or undergoing heavy development, and any minor could potentially contain breaking changes. Limiting the scope to patch updates heavily reduces the chance of our app breaking.
+* The behaviour when installing the dependencies is basically as predictable as possible ([Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)) without the use of npm-shrinkwrap or similar tools.
+
+Don't use wildcards (i.e. `*`, `x`) in the package numbers or keywords like `latest`. They make the versions less clear, and predicting the outcome of installing the dependencies harder.
```
dotcode: "predicting the outcome of installing the dependencies harder" makes my brain hurt. Is it possible to reword, somehow? Maybe using commas, or … I dunno. Not sure how. Work some magic with the English language.

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
...
+[Optional dependencies](https://docs.npmjs.com/files/package.json#optionaldependencies), if present, should also use the same format.
+
+
+### Development dependencies
+
+[Development dependencies](https://docs.npmjs.com/files/package.json#devdependencies) are required during the development, testing, or build process of the app. For example, you may include a test runner, a minifier, a bundler, etc.
+
+These are defined in the `devDependencies` section of the `package.json` file and are not used in production environments.
+
+Version numbers for run-time dependencies are defined using a caret `^` plus a full version number in the MAJOR.MINOR.PATCH form. If the MINOR and PATCH versions are 0, they can be omitted.
```
dotcode: I suggest removing this: "If the PATCH version is 0, it can be omitted."

Should avoid unnecessary exceptions, as it makes for simpler documentation, and in this case will result in less deviation from the norm in the code.

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+
+## Dependency management tools
+
+These are some of the tools that we use to help us manage the dependencies of our projects and alert us of any vulnerabilities in them:
+
+* [Node Security Platform](https://nodesecurity.io/)
+* [Snyk](https://snyk.io/)
+* [Bithound](https://www.bithound.io/)
```
dotcode: I know we've _evaluated_ them, but do we actually use Bithound and NSP? Should clarify. May be helpful for people to understand why we've chosen Snyk …

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
+# Dependency management
+
+The following guide describes how we manage package dependencies in our Node.js-based projects.
+
+
+## Specifying versions of dependencies
+
+As a general rule, we tend to follow a conservative approach when specifying versions for the dependencies of our packages. This helps us ensure consistency between different environments, as we can't know for sure in what environment the app is going to be running, and prevent breakage caused by major or minor updates in dependencies.
```
dotcode: It would be helpful to clarify - here - what we mean by "a conservative approach". Saying something along the lines of "we only allow PATCH version ranges for run-time dependencies, and … etc etc" will help people …

practices/dependency-management.md
```diff
@@ -0,0 +1,104 @@
...
+#### Examples
+
+We do this:
+
+```json
+{
+    "devDependencies": {
+        "jscs": "^2",
+        "mocha": "^3.1",
+        "oxo": "^1.1.1"
```
umaar: Is this meant to be `xo`?

joseluisbolos:
Updated the PR with @dotcode and @umaar's suggestions. Still need to address:
* @alexkilgour's comments
* @dotcode's comments about security tools

joseluisbolos:
Added some clarifications around the use of tilde and caret. @alexkilgour do you mind having a look and see if this makes it clear enough?

joseluisbolos:
@alexkilgour @dotcode Mind having a look at this and see if the explanation is clear enough now?

@dotcode We quickly discussed this at the open-space and people seem to be unsure if the security tools belong here at all, of if they should be here but in a separate document. Would you be ok with merging this as it is, and I will create a new Issue to address specifically the tools section?

dotcode:
Sounds good. Merge away, and thanks :)

PR #150, opened 2017-03-10
add rationale for in-house style & change process

hollsk:
as per recent discussion in The Slack

### DIFF
--- a/technologies/javascript.md
+++ b/technologies/javascript.md
@@ -10,6 +10,7 @@ This document outlines the way we write JavaScript. It's a living styleguide –
   - [KISS](#keep-it-simple)
   - [Performant Code](#performant-code)
 - [Code Style](#code-style)
+  - [Rationale](#rationale)
   - [Linting](#linting)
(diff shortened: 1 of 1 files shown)


jpw:
So many 👍

dotcode:
This is not specific to JavaScript

hollsk:
I was going to put it somewhere else, but nowhere else really jumped out at me as a good location. Any suggestions?

dotcode:
Actually … scratch that previous comment of mine. This rationale section needs to go in every file where we define a style, be it JavaScript or not. I think. What do you reckon?

joseluisbolos:
I feel like quite a lot of the stuff in language/* is actually not specific to those languages and we should have a "code-standards.md" document somewhere.

dotcode:
Yeah, I seem to remember having this feeling before. What do you reckon we commit this, then see if it would help to break some stuff apart in another PR?

hollsk:
Yeah, sounds fair. I created #151 for that and I'll merge this one in.

PR #152, opened 2017-03-14
Changes to code readability section so the message comes across with clarity.

Alicole:
Changes to code readability section so the message comes across with clarity.

### DIFF
--- a/technologies/javascript.md
+++ b/technologies/javascript.md
@@ -71,9 +71,7 @@ function loadConfigFile(filePath, callback) {
 
 ### Optimise For Reading
 
-Code is written once and read many times, using abbreviations or single-character variables saves time in the short term but there's an overhead every time somebody has to read that code.
-
-Reading code is difficult enough at the best of times. Don't make it harder; it's better to have RSI in your over-worked fingers than for all your colleagues to hate you.
+Code tends to be written once and is likely to be mantained by other colleagues over time; so try to make any code produced self descriptive, rather than using code or variable names that are cryptic.
 
 
 ### Modules Over Monoliths
--- a/technologies/javascript.md
+++ b/technologies/javascript.md
@@ -71,7 +71,7 @@ function loadConfigFile(filePath, callback) {
 
 ### Optimise For Reading
 
-Code tends to be written once and is likely to be mantained by other colleagues over time; so try to make any code produced self descriptive, rather than using code or variable names that are cryptic.
+Code tends to be written once and is likely to be maintained by other colleagues over time; so try to make any code produced self descriptive, rather than using code, argument or variable names that are cryptic.
 
 
 ### Modules Over Monoliths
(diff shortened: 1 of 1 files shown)


technologies/javascript.md
```diff
@@ -71,9 +71,7 @@ function loadConfigFile(filePath, callback) {
 
 ### Optimise For Reading
 
-Code is written once and read many times, using abbreviations or single-character variables saves time in the short term but there's an overhead every time somebody has to read that code.
-
-Reading code is difficult enough at the best of times. Don't make it harder; it's better to have RSI in your over-worked fingers than for all your colleagues to hate you.
+Code tends to be written once and is likely to be maintained by other colleagues over time; so try to make any code produced self descriptive, rather than using code, argument or variable names that are cryptic.
```
umaar: Active voice would be good here: https://github.com/springernature/frontend-playbook/blob/master/practices/language.md#plain-english 👍

andrewmee: I also think it would be worth keeping the note on abbreviations and single-character variables since those are very clear examples (and primarily what this section is written to address).

technologies/javascript.md
```diff
@@ -71,8 +71,7 @@ function loadConfigFile(filePath, callback) {
 
 ### Optimise For Reading
 
-Code tends to be written once and is likely to be maintained by other colleagues over time; so try to make any code produced self descriptive, rather than using code, argument or variable names that are cryptic.
-
+Although normally code is written once, it's likely other colleagues maintain your contributions over time; so aim to make any code you produce self descriptive, rather than using cryptic expressions, argument or variable names.
```
umaar: `Although normally` reads funny.

Just an idea, how about:

> You and your peers read code a lot more than you write it. Please ensure you produce easy-to-understand code rather than code which uses cryptic expressions, arguments or variable names.

technologies/javascript.md
```diff
@@ -71,10 +71,7 @@ function loadConfigFile(filePath, callback) {
 
 ### Optimise For Reading
 
-Code is written once and read many times, using abbreviations or single-character variables saves time in the short term but there's an overhead every time somebody has to read that code.
-
-Reading code is difficult enough at the best of times. Don't make it harder; it's better to have RSI in your over-worked fingers than for all your colleagues to hate you.
-
+Although normally code is written once, it's likely other colleagues maintain your contributions over time; so aim to make any code you produce self descriptive, rather than using cryptic expressions, argument or variable names.
```
alexkilgour: This is good, but I think we should keep the explicit example, so maybe:

... rather than using cryptic expressions, argument or variable names. For example, using abbreviations or single-character variables saves time in the short term but there's an overhead every time somebody has to read that code.

technologies/javascript.md
danyalaytekin: - inappropriate semicolon
- `, your colleagues will maintain` would be more accurate
- `self descriptive` requires hyphen

Alicole:
Closing pull request as this section seems to be redundant.

PR #153, opened 2017-03-14
Removing optimize for readability section as is redundant in Javascript documentation of playbook.

Alicole:
Removing optimize for readability section as is redundant in Javascript documentation of playbook.


dotcode:
Looks good.

PR #160, opened 2017-03-24
avoid opening links in new windows

### LINKED ISSUE #103: Detail general approach and advise re: use of "open in new window" links
This should touch on when it may and may not be appropriate to open links in new windows, as well as accessibility concerns.

hollsk:
Resolves #103

I can't really think of a better place for this to go, so I put it in the a11y checklist 💁

### DIFF
--- a/practices/accessibility-checklist.md
+++ b/practices/accessibility-checklist.md
@@ -66,6 +66,7 @@ Be sure to create an issue or submit a pull request if you have any questions, o
    * If form placeholder text is used, is it done in an accessible way (e.g. using HTML5 placeholder attribute)? 
    * If opening a hyperlink in a new window cannot be avoided, has an advanced warning been provided? Does the link have a visible marker indicating its behaviour? 
       * Is that marker also accessible to screenreaders?  
+      
    [WCAG 1, Guideline 10.1](https://www.w3.org/TR/WCAG10/wai-pageauth.html#until-user-agents)  
    [WCAG 2.0, Guideline 3.2.5](https://www.w3.org/TR/WCAG20/#consistent-behavior)  
    [Samurai Errata (guideline 10)](http://www.wcagsamurai.org/erratas/errata-listing/#GL10)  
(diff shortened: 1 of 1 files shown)


practices/accessibility-checklist.md
```diff
@@ -83,6 +83,11 @@ Be sure to create an issue or submit a pull request if you have any questions, o
    [WCAG 1, guideline 14](http://www.w3.org/TR/WCAG10/wai-pageauth.html#gl-facilitate-comprehension)  
    (Also see [WCAG 1 core technique 5: "Comprehension"](http://www.w3.org/TR/WCAG10-CORE-TECHS/#comprehension))
 
+1. **Where opening a hyperlink in a new window cannot be avoided, has an advanced warning been provided?**  
+   Unrequested new windows may cause confusion or disorientation for users with cognitive or visual disabilites. Opening links in new windows is discouraged by both WCAG 1.0 and WCAG 2.0. Does the link have a visible marker indicating its behaviour? Is the marker also accessible to screenreaders?
```
umaar: Would it make sense to open this up to all users? I feel it's a fair statement, e.g.

> Unrequested new windows can cause confusion or disorientation for many users.

hollsk: Mmm, yes. I put it in because the ticket was generated from stakeholders who really, really want all their links to open in new windows, so being able to give very specific reasons to say "no" is useful to us... 

... but by doing that, I've broken with the convention in this file which doesn't mention specific disabilities. I'll go with the convention in the file, the linked resources get more specific. 👍

hollsk: in fact, maybe i should just remove the entire justification

umaar: I think it's ok to remove. The links you provided allow the user to read up on it further if they want to understand the reasoning.

hollsk: Oh, I've just realised that it's already covered by bullet point 13 😆 

That's the trouble with specifications, nobody reads them 😉 

I've combined my additions with point 13 instead

umaar: oops I should've read >.<

hollsk:
thanks @umaar ! 👍

PR #168, opened 2017-04-20
Adds the policy for third party scripts.

joseluisbolos:
Contains an introduction to tag management and the policy for third party scripts, which we have been following for a while but we never had in writing.

### DIFF
--- a/practices/third-party-scripts.md
+++ b/practices/third-party-scripts.md
@@ -1,6 +1,6 @@
 # Third party scripts
 
-Whenever possible, we make use of a tag managers in order to easily manage the scripts being included on a page. We currently use:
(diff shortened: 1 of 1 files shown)


sky-jack:
LGTM 🌮  👍

practices/third-party-scripts.md
```diff
@@ -0,0 +1,22 @@
+# Third party scripts
+
+Whenever possible, we make use of a tag managers in order to easily manage the scripts being included on a page. We currently use:
```
danyalaytekin: `a tag managers`

joseluisbolos: 💯  Sorted.

Issue #120, opened 2016-12-13
Best practices for images


joseluisbolos:
We don't have any section mentioning what to do regarding images, I don't think, so it would be useful to add one mentioning our best practices, e.g.:

* When to use SVG, PNG, JPEG, also vector images vs bitmaps.
* How to optimise images that are going to be commited to a repo (e.g. imageoptim)
* How to optimise images that are going to be server from somewhere else (e.g. [Immagine](https://github.com/springernature/immagine))
* Etc.

joseluisbolos:
This has now been addressed in #170 and #171. Closing.

PR #170, opened 2017-04-24
Add first version of the best practices for images

joseluisbolos:
WIP, more stuff to come. Good to agree on a baseline first in order to avoid bikeshedding.

Starts to address #120.

### DIFF
--- a/practices/images-best-practices.md
+++ b/practices/images-best-practices.md
@@ -1,0 +1,48 @@
+# Best pratices when dealing with images for the Web
+
+In a front-end workflow we typically deal with two kinds of images:
+
+* **Icons, logos, buttons and other graphical elements** that are part of the User Interface of the page. These are usually managed by the frontend, and so they are stored with the rest of resources (templates, CSS, JS). As many of those resources, they may be shared between many different pages.
+
+* **Content images**. These are unique to a page (or a handful of pages). They will be usually provided from a database or a content management system.
+
+Most best practices are common to both categories.
+
+## Formats
+
+### What format to use
+
+* Use SVGs whenever you want to display a vector image.
+* Use JPEGs for photo images, or images with many little details.
+* Use PNGs for graphs and other images with large areas of the same colour. Also for images that require transparency.
+* Don't use GIFs.
+
+### About WebP
+
+We don't encourage the use of the [WebP format](https://developers.google.com/speed/webp/) as support for it is still not widespread. [MS is not planning to add support to Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/webpimageformat/) at the moment. Firefox is currently working in adding support for it, so we may reconsider after that.
+
+In the latest years, new encoders like [Mozilla's mozjpeg](https://github.com/mozilla/mozjpeg) or [Google's Guetzli](https://github.com/google/guetzli) have appeared that can create better optimised JPEGs. Therefore WebP is not as advantageous as it used to be.
+
+### JPEG images
+
+When encoding JPEG images:
+
+* Use progressive encoding. Progressive JPEGs provide a better user experience and, most of the time, are smaller than baseline. See this [yeoman issue](https://github.com/yeoman/yeoman/issues/810) for further details.
+* Use [4:2:0](https://en.wikipedia.org/wiki/Chroma_subsampling#4:2:0) [chroma subsampling](https://en.wikipedia.org/wiki/Chroma_subsampling) whenever possible. This encodes the colour information at half the resolution of the luma, which may result in a perceptible loss of colour accuracy. This will usually generate much smaller images, so we suggest to create both of them and compare them visually for any loss of quality.
+
+### PNG images
+
+Always use the maximum compression level (9).
+
+## Optimisation
+
+* Make sure that you save images using an sRGB ICC profile. Although modern browsers support images with ICCv2 profiles like AdobeRGB and even Profoto, colour space issues are extremely difficult to debug. Saving the images as sRGB and removing the colour profile is encouraged.
+* Stip the image from all metadata. This includes:
+    * EXIF data
+    * Copyright information (unless required)
+    * GPS data
+    * Colour profiles
+    * Thumbnails in JPEGs
+    * Comments
+    * Etc.
+* Set appropiate cache headers for the images.
--- a/practices/images-best-practices.md
+++ b/practices/images-best-practices.md
@@ -2,7 +2,7 @@
 
 In a front-end workflow we typically deal with two kinds of images:
 
-* **Icons, logos, buttons and other graphical elements** that are part of the User Interface of the page. These are usually managed by the frontend, and so they are stored with the rest of resources (templates, CSS, JS). As many of those resources, they may be shared between many different pages.
+* **Icons, logos, buttons and other graphical elements** that are part of the User Interface of the page. These are usually managed by the frontend, and so they are stored with the rest of resources (templates, CSS, JS). Like those resources, they may be shared between many different pages.
 
 * **Content images**. These are unique to a page (or a handful of pages). They will be usually provided from a database or a content management system.
 
@@ -21,7 +21,7 @@ Most best practices are common to both categories.
 
 We don't encourage the use of the [WebP format](https://developers.google.com/speed/webp/) as support for it is still not widespread. [MS is not planning to add support to Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/webpimageformat/) at the moment. Firefox is currently working in adding support for it, so we may reconsider after that.
 
-In the latest years, new encoders like [Mozilla's mozjpeg](https://github.com/mozilla/mozjpeg) or [Google's Guetzli](https://github.com/google/guetzli) have appeared that can create better optimised JPEGs. Therefore WebP is not as advantageous as it used to be.
+Recently, new encoders like [Mozilla's mozjpeg](https://github.com/mozilla/mozjpeg) or [Google's Guetzli](https://github.com/google/guetzli) have appeared that can create better optimised JPEGs. Therefore WebP is not as advantageous as it used to be.
 
 ### JPEG images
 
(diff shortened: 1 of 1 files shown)


practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+* Use [4:2:0](https://en.wikipedia.org/wiki/Chroma_subsampling#4:2:0) [chroma subsampling](https://en.wikipedia.org/wiki/Chroma_subsampling) whenever possible. This encodes the colour information at half the resolution of the luma, which may result in a perceptible loss of colour accuracy. This will usually generate much smaller images, so we suggest to create both of them and compare them visually for any loss of quality.
+
+### PNG images
+
+Always use the maximum compression level (9).
+
+## Optimisation
+
+* Make sure that you save images using an sRGB ICC profile. Although modern browsers support images with ICCv2 profiles like AdobeRGB and even Profoto, colour space issues are extremely difficult to debug. Saving the images as sRGB and removing the colour profile is encouraged.
+* Stip the image from all metadata. This includes:
```
danyalaytekin: Should this say "Strip metadata from the image"?

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
+# Best practices when dealing with images for the Web
+
+In a front-end workflow we typically deal with two kinds of images:
+
+* **Icons, logos, buttons and other graphical elements** that are part of the User Interface of the page. These are usually managed by the frontend, and so they are stored with the rest of resources (templates, CSS, JS). As many of those resources, they may be shared between many different pages.
```
dotcode: > As many of those resources

Suggest changing to "Like those resources …"

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+
+* **Content images**. These are unique to a page (or a handful of pages). They will be usually provided from a database or a content management system.
+
+Most best practices are common to both categories.
+
+## Formats
+
+### What format to use
+
+* Use SVGs whenever you want to display a vector image.
```
dotcode: This is equivalent to saying "use PNGs whenever you want to display a bitmap image". Instead describe like you have JPEGs and PNGs: what is the _type_ that requires the use of a vector image?

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+* Use JPEGs for photo images, or images with many little details.
+* Use PNGs for graphs and other images with large areas of the same colour. Also for images that require transparency.
+* Don't use GIFs.
+
+### About WebP
+
+We don't encourage the use of the [WebP format](https://developers.google.com/speed/webp/) as support for it is still not widespread. [MS is not planning to add support to Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/webpimageformat/) at the moment. Firefox is currently working in adding support for it, so we may reconsider after that.
+
+In the latest years, new encoders like [Mozilla's mozjpeg](https://github.com/mozilla/mozjpeg) or [Google's Guetzli](https://github.com/google/guetzli) have appeared that can create better optimised JPEGs. Therefore WebP is not as advantageous as it used to be.
```
dotcode: > In the latest years

Suggest changing to "Recently"

practices/images-best-practices.md
joseluisbolos: What about:

> "Use SVGs for icons, logos, and other vector images."

dotcode: > and other vector images

"and other images with sharply defined edges and areas of …"

I don't know. Something along those lines. Refrain from saying "use vector images where you should use vector images".

joseluisbolos: Hmm... I see what you mean. When I wrote this I was concerned about the source of the image though, as that usually is not under our control: we can only use SVG if we have the original image in a vector format (AI or whatever) as there's no (easy) way back from a bitmap image to a vector one.

So, at risk of being redundant, what I wanted to say is "if you're lucky enough to have a vector image and the original vector format (AI or whatever) then just use that one. Don't convert it to a bitmap-based format."

joseluisbolos:
All issues addressed, still looking for a good wording for the SVG section.

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+Most best practices are common to both categories.
+
+## Formats
+
+### What format to use
+
+* Use SVGs whenever you want to display a vector image.
+* Use JPEGs for photo images, or images with many little details.
+* Use PNGs for graphs and other images with large areas of the same colour. Also for images that require transparency.
+* Don't use GIFs.
```
umaar: Maybe an explanation as to why? What are the alternatives e.g

While gifs might be terrible for performance I think there are still a few use cases for them.

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+
+### About WebP
+
+We don't encourage the use of the [WebP format](https://developers.google.com/speed/webp/) as support for it is still not widespread. [MS is not planning to add support to Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/webpimageformat/) at the moment. Firefox is currently working in adding support for it, so we may reconsider after that.
```
umaar: Does 'not encouraging' something mean to discourage it?

Anyway, if global support for webp is [over 70%](http://caniuse.com/#search=webp) - not sure if this is a good recommendation since the appropriate image format can be served to supporting clients via content negotiation.

practices/images-best-practices.md
```diff
@@ -0,0 +1,48 @@
...
+
+Recently, new encoders like [Mozilla's mozjpeg](https://github.com/mozilla/mozjpeg) or [Google's Guetzli](https://github.com/google/guetzli) have appeared that can create better optimised JPEGs. Therefore WebP is not as advantageous as it used to be.
```
umaar: Are we certain that the optimisation of JPEGs with these tools is superior to WebP? As in, does Guetzli/mozjpeg JPEGs result in smaller sized WebP images?

practices/images-best-practices.md
joseluisbolos: I'm trying to document our current practices. Are we currently using WebP anywhere?

practices/images-best-practices.md
joseluisbolos: No, JPEG is worse than lossy WebP, but with this new encoders, the difference is not as big as it used to be. Therefore WebP is not as advantageous as it used to be.

practices/images-best-practices.md
joseluisbolos: What would be an acceptable use case for GIFs?

practices/images-best-practices.md
umaar: But:

1. The main title of this document is "Best practices when dealing with images for the Web". Consider that the playbook is also a community resource so it should be applicable for community members who might be reading this.  
2. @dotcode explained the playbook is "What we aspire to do".  

I don't know if we're using WebP anywhere, I haven't seen it in use, but my point is, shouldn't we be recommending formats which deliver the best user experience?

practices/images-best-practices.md
umaar: If you agree with the earlier point on reconsidering the WebP text, I would remove:

> Therefore WebP is not as advantageous as it used to be.

Since I'm not sure what it adds to anything?

practices/images-best-practices.md
umaar: If you are given an animated gif to display on a page, and autoplaying videos aren't supported on browsers which you're meant to support - just an example.

joseluisbolos: Yep, but as far as I can tell, we don't use animated gifs (or autoplay videos) at Springer Nature. And I'm just trying to document what we do.

practices/images-best-practices.md
joseluisbolos: I don't personally aspire to use WebP, I don't think that it's a good solution, for many reasons that don't deserve to be written on a first draft, as I'm mostly interested in writing down other stuff that I have in the pipeline. I'm mostly trying to document what we're currently doing.

So I guess that this bows down to:
* Are we using WebP anywhere, and why?
* If not, are we going to use it soon, and why?

Even if the document is _partly_ aspirational, I think that saying "you should be using WebP" when we're not using it ourselves, nor planning to use it, is misleading.

But I feel that that discussion don't belong to this PR, so seeing as this is a contentious point I'm removing the GIF and WebP sections altogether so we can get this merged.

umaar: Hmm not sure about this. It we didn't concatenate our JavaScript, we wouldn't write in the playbook "Don't concatenate your JavaScript". But it feels like that's what this text is doing?

If you purely want to state _What we currently do_ and exclude _What is a best practice applicable for everyone_, shouldn't we clearly differentiate between them? 

Also if you're discouraging it's use (which is how I read the text) - then I think it _is_ worth backing up in this PR. I like to share the frontend playbook with other members in the community, and at first glance, reading this document feels like we're recommending folks to _not_ use WebP - it's not entirely clear that you're simply stating, "we, Springer Nature, do not currently use WebP".

practices/images-best-practices.md
```diff
@@ -0,0 +1,41 @@
+# Best practices when dealing with images for the Web
+
+In a front-end workflow we typically deal with two kinds of images:
+
+* **Icons, logos, buttons and other graphical elements** that are part of the User Interface of the page. These are usually managed by the frontend, and so they are stored with the rest of resources (templates, CSS, JS). Like those resources, they may be shared between many different pages.
```
danyalaytekin: Are logos always UI?
Also, maybe `the rest of the resources` or `the other resources`.

practices/images-best-practices.md
umaar: Got it, I think I'm reading this as something applicable for everyone (rather than just "Here's what we currently do").

practices/images-best-practices.md
```diff
@@ -0,0 +1,41 @@
...
+
+Most best practices are common to both categories.
+
+## Formats
+
+### What format to use
+
+* Use SVGs whenever you want to display a vector image.
+* Use JPEGs for photo images, or images with many little details.
+* Use PNGs for graphs and other images with large areas of the same colour. Also for images that require transparency.
```
danyalaytekin: I think I saw some nostalgia for GIFs earlier. You could mention PNG8s have superseded GIFs (if not animated) and are smaller.

practices/images-best-practices.md
joseluisbolos: Not sure what you mean with "the rest of the resources or the other resources".

practices/images-best-practices.md
joseluisbolos: Not really because I want this PR to be merged, but if you fancy you can add that bit after this one has been merged.

practices/images-best-practices.md
danyalaytekin: Oh, those were two separate suggestions. Grey backgrounds. Small grammer change innit?

PR #173, opened 2017-04-27
Make headings sentence case

danyalaytekin:
This PR standardises the case of headings which were previously a mix of sentence case and word case.  I've chosen sentence case, which was the case used in the root's README, so `Some Heading` has become `Some heading`, hopefully everywhere.  But say if you have a preference for another case.

### DIFF
--- a/practices/code-review.md
+++ b/practices/code-review.md
@@ -1,4 +1,4 @@
-# Code Reviews
+# Code reviews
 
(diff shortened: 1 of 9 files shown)


danyalaytekin:
A relevant discussion :) https://github.com/pa11y/pa11y.github.io/issues/39

Issue #177, opened 2017-06-22
Suggestion to make peer review FED process more inclusive


Alicole:
It might be sensible to make as small tweak to the wording on the PR playbook docs, on following line in Why section to make it more inclusive to wider dev community, as sometimes not just FEDs change templates and code; from, "Ensure that Front-End Developers within each team are coding to the same guidelines." To: "Ensure that any Developers, within each team, who are changing front-end templates or code are conforming to the same guidelines".

Please see: https://github.com/springernature/frontend-playbook/blob/master/practices/code-review.md

jpw:
Closing for now, keeping an eye on the situation.

PR #179, opened 2017-06-22
playbook #177: Suggestion to make peer review FED process more inclusive

Alicole:
Added line to explain benefits of PR to a member not necessarily from from front end.


jpw:
Not necessary and I would recommend not rattling that particular cage.

danyalaytekin:
Agreed.

Issue #180, opened 2017-06-26
Add a "performance" section


hollsk:
I propose a "performance" section to contain the following existing files:

* advertising
* images
* graded browser support
* progressive enhancement

Is this idea awesome Y/N?

jpw:
Is this idea awesome Y/~~N~~?

davidpauljunior:
![y](https://user-images.githubusercontent.com/1489314/27583799-8fe1c98a-5b2d-11e7-98a6-6fedc1b227a5.png)

davidpauljunior:
I'm not sure I'd instinctively think about graded browser support and progressive enhancement as being under the umbrella of 'performance'.

hollsk:
Yeah, i had the same thought @davidpauljunior - but I couldn't think of anything more appropriate. 

* They don't belong in `markup` or `accessibility` 
* Both are distinct (& related) technical approaches, but they're not really `architecture`, exactly

Can you think of a category they could be contained by?

jpw:
general concencus here to leave everything in the root unless > 1 related thing, then folder

don't know if this answers the question lol

hollsk:
As per catchup:

* everyone agrees that GBS and PE don't strictly belong in performance
* we agreed that it's fine to leave pages uncategorised if there's no logical grouping
* general decision made to leave everything in the root unless > 1 related thing, then folder

Issue #143, opened 2017-03-02
Review the "Dependency management tools" section


josebolos:
We added a section about tools for managing Node.js dependencies to our [Dependency management page](https://github.com/springernature/frontend-playbook/blob/master/practices/dependency-management.md#dependency-management-tools).

Maybe it's not the best place to contain that information and we should have a separate page for tools.

danyalaytekin:
Waiting for a volunteer.

hollsk:
I was looking at the semver page earlier and thinking it would be good to have an entire section for Node that included semver and dep management. Could also contain the resolution to #121 once we do that. 

What say you?

joseluisbolos:
I like it. What would that page include?

* Dep management?
* Tools?
* ...
* Profit?

hollsk:
Yes. Tools and dep management for now. I thought semver too, but jon has just moved the semver page under the git folder, so maybe not 😁 

Could also have a bit about how we manage our NPM stuff (aspirationally, i mean, rather than currently 😂  )

joseluisbolos:
#121 to be addressed as part of this. So we have:

* Dep management
* NPM Publising and namespacing
* Tools

josebolos:
Addressed in #183.

PR #181, opened 2017-06-26
adjust ALL the things

hollsk:
BEHOLD! My proposal!

* JS, CSS, and markup get a directory of their own under the root, each of which contains a `house-style.md` along with anything other files related to their overarching theme
* Some of these directories will have lots of files in, some won't. NBD
* Accessibility, git, and general writing guides all get their own directories under root

This leaves the "practices" directory looking like a chum bucket of unrelated concepts 😷 but it's ok because:

* I've created an issue #180 to contain anything perf related
* We already have another issue #143 for general node concepts

What say you?

(here's the proposed tree which is easier to look at than the commit diff: https://github.com/springernature/frontend-playbook/tree/acdb516ed13f3610687ae0cc766a4a8b98177ff1 )

(and here's the issue that prompted all this in the first plaice 🐟 https://github.com/springernature/frontend-open-space/issues/94 [private repo])


davidpauljunior:
This looks like a good approach to me.  I like that fact that you could pass someone a url that encompasses a language entirely.

jpw:
Nice! Did you check the links didn't break?

hollsk:
I went through every single file and updated every single link that needed updated, but obviously I'd appreciate people clicking about a bit just to make sure 😁

hollsk:
Or, you know, I could just merge and if somebody sees something broken they can damn well fix it themselves 😇

dotcode:
Bit late to the party, but: excellent work @hollsk, top job.

Issue #121, opened 2016-12-13
Namespacing node projects


joseluisbolos:
Not namespacing stuff is a very bad practice in general. That that npm and most other tools have proper support for namespacing, we should aim to namespace every new project node-based project that we release (e.g. `@springernature/potato` instead of `potato`). 

Recommendation to be added to the playbook.

joseluisbolos:
Will be addressed as part of #143

PR #183, opened 2017-07-04
Update node practices document

josebolos:
Renames the file and adds "publishing to npm" and snyk sections. This addresses #121 and #143.


josebolos:
I see. The choice of words in this particular sentence was intentional. I was using `must` as specified in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt), which is what we usually do for all our docs, implying that adding the project to Snyk is not optional.

If people think that it's not clear that way I'll change it. Let's see what others think.

hollsk:
"must", "should", "required" etc is consistent with our approach to language throughout

we should really add a link to RFC 2119 somewhere prominent, say on the README itself, or in the house style guides 

it's linked inside individual pages but a bit of tidying would be nice (another PR though)

i still disagree with lowercasing those keywords, fwiw - if we'd continued to uppercase them then it would have been more obvious that this was a requirement level keyword

hollsk:
in fact, it might also be good to add a note about using requirement level keywords to the language guide itself

josebolos:
Addressed in #184.

PR #184, opened 2017-07-04
adjust language section to cover technical writing

hollsk:
* Contents reordered to put writing style at the top
* New "non-discriminatory language" section that bundles the related sections
* New "technical writing" section
* Misc consistency fixes to capitalisation, we do/don't do, formatting and links 

Technical writing section is short. There's not really much I can say here that's not already covered in the existing plain english section, so i've made it explicit that concision and structure is better than chatty in technical contexts, and added a link to RFC 2119.

Work to be completed in another PR(s):
* Still not happy with the ordering of this file so I need to take another pass at it
* Need to add 2119 references elsewhere too

### DIFF
--- a/writing/language.md
+++ b/writing/language.md
@@ -8,7 +8,7 @@
     * [Unacceptable language](#unacceptable-language)
     * [Examples](#examples)
         * [Genderless language](#genderless-language)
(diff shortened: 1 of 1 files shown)


Alicole:
Looks fine to me -- any other comments?

PR #191, opened 2017-08-16
How we write CSS

alexkilgour:
* Add a `How we write CSS` page detailing our componentised approached to writing CSS
* Remove mosaic and OOCSS pages (move content to shunter-mosaic wiki)
* Update BEM and house style pages to reflect new approach


css/bem-css.md
```diff
@@ -38,58 +40,58 @@ We do this:
...
 
 We do this:
 ```css
-.alert-box
-.button
+.c-alert-box
+.c-button
 ```
 
 **Element**: styles that only apply to children of a block. Elements can also be blocks themselves. Class name is a concatenation of the block name, two underscores and the element name. Examples:
```
benjclark: `Elements can also be blocks themselves`
Think this might be a bit misleading. We wouldn't accept someone treating a BEM element as a block and then creating an element on the back of that e.g. `c-alert-box__close__link`.

benjclark: "Elements can be converted in to blocks if needed" perhaps

alexkilgour: Agreed that is confusing. I'll just take it out as I think it is explained how to tackle this further down at line 76

benjclark: Yeah good shout 👍

css/bem-css.md
```diff
@@ -38,58 +40,58 @@ We do this:
...
 
 **Modifier**: override or extend the base styles of a block or element with modifier styles. Class name is a concatenation of the block (or element) name, two hyphens and the modifier name. Examples:
-- `.alert-box--success`
-- `.button--expanded`
+- `.c-alert-box--success`
+- `.c-button--expanded`
 
 ## BEM best practices
 
 Don't `@extend` block modifiers with the block base, always require that the base/unmodified class is present
```
davidpauljunior: Should this be _without_ the block base?

css/bem-css.md
```diff
@@ -105,12 +107,12 @@ If you are using a CSS pre-processor, excessive nesting can be avoided by using
 Targeting a BEM element/modifier from within a nested block rule (see above) should be done with caution. If your selectors contain a large number of declarations then you should consider moving the element out of the nested rule to make them more readable, and easier to follow. Always adhere to the [KISS](https://en.wikipedia.org/wiki/KISS_principle) principle. You can target a modifier as follows, whilst keeping the element separate for better readability:
 
 ```scss
-.block {
+.c-block {
     &--modifier { // compiles to .block--modifier
```
davidpauljunior: Just need to update the `.block` references in the comments to `.c-block`

css/bem-css.md
```diff
@@ -105,12 +107,12 @@ If you are using a CSS pre-processor, excessive nesting can be avoided by using
...
         text-align: center;
     }
 }
-.block__element {
+.c-block__element {
     color: red;
 
     &--modifier { // compiles to .block__element--modifier
```
davidpauljunior: Same as above

css/bem-css.md
alexkilgour: good catch

css/bem-css.md
alexkilgour: Actually think it's a bit ambiguous, could be either depending on how you read it. Maybe change to `Don't @extend block modifiers on the block base`?

css/house-style.md
```diff
@@ -188,6 +183,8 @@ One exception to these rules is targeting a BEM element from within a nested blo
 
 ## The Rules
 
+Projects should use [StyleLint](https://github.com/stylelint/stylelint) to enforce these rules. You can find _an example_ config file in the [config directory](config).
+
 ### Indentation
 
 We indent our css using single tabs, not spaces. You can convert characters automatically in most editors, and you're advised to do this.
```
davidpauljunior: CSS should be capitalised here for consistency

alexkilgour: 👍

css/house-style.md
```diff
@@ -691,33 +688,33 @@ a { color: pink; }
...
-.ma4 { margin: 4px; }
-.ma6 { margin: 6px; }
-.ma10 { margin: 10px; }
-.ma15 { margin: 15px; }
-.ma20 { margin: 20px; }
-.ma30 { margin: 30px; }
-.ma40 { margin: 40px; }
-.ma50 { margin: 50px; }
-.ma70 { margin: 70px; }
+.c-ma0 { margin: 0; }
```
davidpauljunior: Out of curiosity, does this go against the previously described rules around naming?
E.g. we say we use `c-button` not `c-btn`.  But here the example is `c-ma0`, rather than `c-margin0`

css/bem-css.md
davidpauljunior: I think that's more confusing 😆 
Actually I'm confused by the use of `@extend`.  Are we actually talking about SASS `@extend`?  I thought this was just about saying, you can't have a modifier without a base class to modify.

css/how-we-write-css.md
```diff
@@ -0,0 +1,172 @@
...
+* Contains _only_ variable definitions
+
+#### 20 Functions
+* Contains _only_ function definitions
+
+#### 30 Mixins
+* Can include _private_ functions and variables (exist at this level)
+
+#### 40 Base
+* Should avoid type selectors where possible
```
davidpauljunior: I thought _base_ was primarily aimed at type selectors?
e.g. where you'd find your `html, body, p, ul, ol, a` etc defaults.

css/bem-css.md
alexkilgour: I don't know, this was all already here! 😆

davidpauljunior: I know sorry, I'm reviewing old stuff 😳

alexkilgour: `Don't directly extend the base block element with a modifier, always require that the unmodified class is present`?

css/how-we-write-css.md
alexkilgour: ah yes, should be `class`, not `type`

alexkilgour: as in avoid `class` selectors where possible

css/house-style.md
alexkilgour: yes, this should be updated. good spot.

css/how-we-write-css.md
davidpauljunior: oh haha, that makes sense.

PR #195, opened 2017-08-17
JavaScript architecture recommendations

umaar:


javascript/house-style.md
```diff
@@ -776,6 +782,143 @@ In-browser you should not add your `'use strict';` to the top of the file; inste
...
+exports default { init };
+```
+
+### DOM Binding
+
+If you have a heavy JavaScript application and you can justify the performance penalty your users will pay for the download + execution of your script, then a JavaScript framework might make sense. Consider something small like [Preact](https://github.com/developit/preact). Using a JavaScript framework can handle DOM binding efficiently.
+
+For the majority of websites, especially those which may by viewed on low powered devices, you should minimise the JavaScript you send down the wire and implement DOM binding yourself. For simple use cases, manually cherry picking elements out of the DOM and reading/setting attributes on them is a reasonable approach.
 
 ### Directory structure
```
alexkilgour: This is really nice. I'm just wondering if some of the directions within `Directory structure` are now redundant - so the stuff about binding to the `SN` namespace, and the stuff about jQuery as we are trying to move away from jQuery

zanonnicola:
It would be also good promoting to write modules to be "tree-shanking" compatible (Webpack, RollUp). 

Do:

```js
// module.js

const makeMeASandwich = (a, b) => {
  return `make sandwich: ${a}, ${b} `;
}
const makeMeAnotherSandwich = (a, b) => {
  return `make sandwich: ${a}, ${b} `;
}

export {makeMeASandwich, makeMeASandwich};
```

Don't

```js
// module.js

const sandwich = {
	makeMeASandwich(a, b) {
		return `make sandwich: ${a}, ${b}`;
	},
	makeMeAnotherSandwich(a, b) {
		return `make sandwich: ${a}, ${b}`;
	}
};

export default sandwich;
```

umaar:
@zanonnicola done! Good suggestion

javascript/house-style.md
umaar: True! I removed the legacy points which don't really apply

alexkilgour:
👍

PR #204, opened 2017-11-15
adding Microsoft accessibility image

dotcode:


josebolos:
URL is still broken, and it's also missing the https://

PR #206, opened 2017-11-15
Add section on methods for binding JS modules to DOM elements

andrewmee:
Attempting to document our convention for doing this.

I believe this is accurate as per how we do things across most of our big applications here at Springer Nature (mosaic-style apps, BMC, and SpringerLink), but open to be corrected if anyone has a different view on this or an alternative they'd like us to move to.

### DIFF
--- a/javascript/house-style.md
+++ b/javascript/house-style.md
@@ -960,7 +960,7 @@ We don't do this:
 <nav id="popout" class="c-popout-nav c-popout-nav--dark">...</nav>
 ```
 
-As demonstrated above, we use `data-component` attributes to label the "main" element that is being bound by the JavaScript, and also any additional elements used by the module which can't otherwise be unabiguously selected.  In instaces where child elements can be inferred without using additional attributes &mdash; for example a module which is expected to contain just a single input &mdash; then it may be acceptable to use other selectors such as type selectors, however care should be taken.
+As demonstrated above, we use `data-component` attributes to label any elements that are being bound by the JavaScript.  In instances where other related elements can be inferred without using additional data attributes &mdash; for example a module which is expected to contain just a single input &mdash; then it may be acceptable to use other selectors such as type selectors, however care should be taken.
 
 Initialising a module can then be achieved simply by passing in one of more elements that match these selectors;
 
--- a/javascript/house-style.md
+++ b/javascript/house-style.md
@@ -962,7 +962,35 @@ We don't do this:
 
 As demonstrated above, we use `data-component` attributes to label the "main" element that is being bound by the JavaScript, and also any additional elements used by the module which can't otherwise be unabiguously selected.  In instaces where child elements can be inferred without using additional attributes &mdash; for example a module which is expected to contain just a single input &mdash; then it may be acceptable to use other selectors such as type selectors, however care should be taken.
 
-The selector(s) based on these data-attributes may be defined as a default selector within the JavaScript module, which is able to be overridden in the `init` call.  Alternatively for simpler cases you may choose to keep the selector outside of the module itself, and instead initialise the module with a collection of elements.
+Initialising a module can then be achieved simply by passing in one of more elements that match these selectors;
+
+```js
+const moduleInstances = document.querySelectorAll('[data-component="simple-module"]');
+if (moduleInstances.length > 0) {
+    mySimpleModule.init(moduleInstances);
+}
+```
+
+Alternatively for more complex modules that need to be bound to multiple DOM elements, it may make more sense to have the module itself contain a set of default selectors which can then be (optionally) overridden in the init call;
+
+```js
+function init(settings) {
+    const defaults = {
+        wrapper: '[data-component="complex-module"]',
+        close: '[data-component="complex-module-close"]'
+    };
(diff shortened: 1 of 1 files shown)


javascript/house-style.md
```diff
@@ -939,6 +939,33 @@ exports default { init };
...
+
+```html
+<nav class="c-popout-nav c-popout-nav--dark" data-popout>...</nav>
+
+<nav class="js-popout c-popout-nav c-popout-nav--dark">...</nav>
+
+<nav id="popout" class="c-popout-nav c-popout-nav--dark">...</nav>
+```
+
+As demonstrated above, we use `data-component` attributes to label the "main" element that is being bound by the JavaScript, and also any additional elements used by the module which can't otherwise be unabiguously selected.  In instaces where child elements can be inferred without using additional attributes &mdash; for example a module which is expected to contain just a single input &mdash; then it may be acceptable to use other selectors such as type selectors, however care should be taken.
```
alexkilgour: "also any additional elements used by the module which can't otherwise be unabiguously selected"

On SpringerLink we use `data-role="name"` for these. The module is always bound to a `data-component` and any child can be given a `data-role`

javascript/house-style.md
```diff
@@ -939,6 +939,33 @@ exports default { init };
...
+
+The selector(s) based on these data-attributes may be defined as a default selector within the JavaScript module, which is able to be overridden in the `init` call.  Alternatively for simpler cases you may choose to keep the selector outside of the module itself, and instead initialise the module with a collection of elements.
+
```
alexkilgour: I don't actually understand this paragraph! But that is probably saying more about me 😁

zanonnicola:
This is great for end to end tests as well. Using `data-component` for JS hooks makes your UI tests resilient to change. If you add a new feature or refactor your HTML + CSS you won't have to modify your tests.

javascript/house-style.md
andrewmee: Ah interesting.  Yeah I thought this might be a contentious bit — it's something @davidpauljunior and I have talked about whether to reuse `data-component` or something else.  David had suggested another option was to re-use the component name e.g. `data-component="popout"` then children like `data-popout="close"`.

Personally I favour have more standardised attribute names, but using a single alternate one like `data-role` may be a middle-ground solution.

javascript/house-style.md
dotcode: Would examples help?

andrewmee: I might be better off rephrasing this and rearranging so that the simple-case is the first thing mentioned.  And yeah perhaps examples would help make both clearer!

javascript/house-style.md
benjclark: Do we need to write about main and additional elements? I'm wondering if that's irrelevant really and maybe certain modules won't have a main hook but will have multiple hooks. Personally I like the idea that every js hook follows the same `data-component` convention, perhaps with a prefix to associate it with a js module, e.g. `data-component="popout"`, `data-component="popout-close"`, `data-component="popout-totally-amazing-parallax-carousel-figit-spinner"`

andrewmee: That's a good point that the idea of a "main" element may not make much sense, depending on what the module is doing.

If we went this way I could change that paragraph to;

>As demonstrated above, we use `data-component` attributes to label any elements that are being bound by the JavaScript.  In instances where other related elements can be inferred without using additional data attributes &mdash; for example a module which is expected to contain just a single input &mdash; then it may be acceptable to use other selectors such as type selectors, however care should be taken.

Makes it quite a lot simpler.

alexkilgour: Agreed, this does simplify things and allows for more complex use cases 👍

javascript/house-style.md
```diff
@@ -939,6 +939,61 @@ exports default { init };
...
+module.exports = {init};
+```
+```js
+myComplexModule.init({
+      wrapper: '[data-component="renamed-complex-module"]',
+      close: '[data-component="renamed-complex-module-close"]'
+});
+```
+
```
alexkilgour: This is really good, thanks 💯

javascript/house-style.md
sonniesedge: Labelling things as `data-component="whatever" data-component="another"` and then passing them into a module as needed looks great. The naming convention is flexible and allows the naming itself to imply child-parent relationships (e.g. `data-component="whatever"` and 'data-component="whatever-child"').

Like @benjclark says, there's going to be plenty of cases where you don't have a "main" component, so the removal of main makes sense to me.

javascript/house-style.md
```diff
@@ -939,6 +939,61 @@ exports default { init };
...
+Initialising a module can then be achieved simply by passing in one of more elements that match these selectors;
+
+```js
+const moduleInstances = document.querySelectorAll('[data-component="simple-module"]');
+if (moduleInstances.length > 0) {
+    mySimpleModule.init(moduleInstances);
+}
+```
+
+Alternatively for more complex modules that need to be bound to multiple DOM elements, it may make more sense to have the module itself contain a set of default selectors which can then be (optionally) overridden in the init call;
```
sonniesedge: 👍

PR #207, opened 2017-11-21
 expands upon polyfill policy

jpw:


javascript/house-style.md
```diff
@@ -1006,14 +1006,18 @@ Client-side templating is typically to be avoided - e.g. in the example of an AJ
...
+
 If you utilise a Polyfill, the JavaScript must meet these conditions:
 
   - It must _not_ override native browser functionality
   - It must match native browser functionality exactly (where possible)
   - It can be either third-party, or written in-house
-  - Must be installed via NPM
+  - Must be installed via NPM (or dynamically via [polyfill.io](https://polyfill.io/v2/docs/), at your discretion).
+
+Note that `polyfill.io` adds one additional blocking HTTP request, and if loading asynchronously one must be careful to manage loading order of scripts.  For these reasons it is preferred to bundle polyfills with your main application code.
```
andrewmee: I'm not sure I agree with this addition.  We found polyfill.io gave us much better results than our own bundled polyfills and with a lot less hassle!

jpw: Thanks for having a look Andrew.
Hrms OK, so how did using polyfill.io give you "better results"?
And is there anything you dispute with the performance aspects of using it, or... something else?

andrewmee: Sorry, I should have been clearer with the bit I found problematic; it is the "For these reasons it is preferred to bundle polyfills with your main application code."

Previously on BMC we'd bundled a bunch of polyfills in our vendor.js file, but we found that actually polyfill.io could provide the same 'coverage' with a greatly reduced payload size, and of course while moving those polyfills off our books vis-a-vis dependencies and build steps.

So I'm not sure if I wouldn't be more inclined to recommend taking the extra HTTP request in exchange for these gains.

jpw: It's interesting that their polyfills were so much slimmer!
Anyway, re. performance, it's the blockingness of the HTTP request that irks especially and it seems quite a hit compared to the day-to-day of having those extra things in the build and probably just forgotten about.  However, it's your team that has been living with this on the day-to-day, so am happy to reword to be less dogmatic.

josebolos: Is it missing a "the" before "loading order of scripts", as in:

"if loading asynchronously one must be careful to manage _the_ loading order of scripts"

jpw: I don't think so, but with is more clear, so will add.

jpw: Updated!

jpw:
Thanks guys!

PR #208, opened 2017-11-29
Refreshes data attributes approach

jpw:
Also auto-generated a TOC thereby adding a few missing things and making some headings consistent.


jpw:
If we feel the TOC is too long I can stop it generating entries for the last dimension

jpw:
Thanks for looking folks!

PR #209, opened 2017-11-30
Add section dealing with classification of built assets/dependencies

andrewmee:
There has been some discussion recently about whether some dependencies should be classified as run-time dependencies or devDependencies, given they're being run through gulp or some other build tool.

Adding this section to outline an approach for dealing with these potential grey-area assets.

I think it's important that we do create a common documented approach for this, as otherwise there can be confusion which leads to an illogical splitting between the two dependency "buckets" within `package.json` files.


jpw:
This is quite involved;
1. Our [current practice](https://github.com/springernature/frontend-playbook/blob/master/practices/managing-node-projects.md#run-time-dependencies) - for node projects at least - is to `~x.x.x` deps, and `^x.x.x` dev deps. This means dev deps can get minor hikes without us noticing.
1. this is probably not a big deal for node projects as dev deps are unlikely to get bundled up and end up in production. However, a browser project where e.g. a npm module that doesn't _need_ its source in production (and so is a dev dep) may get transpiled or whatever and end up in an asset bundle on production. In that case, a minor version hike caused by a `^` could be a nasty surprise in production.
1. it will probably be confusing to have different standards for node vs browser projects, so we'd like to keep it consistent

Solutions;
1. Different standards for browser & node projects. Quite confusing, especially where both are combined.
1. Tilde all deps. Simple, but loses the auto minor version hiking for dev deps.
1. Do as per this PR proposal. Upside, optimises for humans not machines (usually a good idea) i.e. a developer can see clearly what is "really" a prod or dev dep. Downsides, longer build times as larger dep trees are built, plus unused code (that won't ever be actually _run_) ending up in prod. And we have to remember to do it (I _suspect_ we haven't been that rigorous following the current practice anyway).
1. Do nothing.

Please feel free to chip in folks and correct me where I've got it rong!

Alicole:
The GOCD build step can be quite timely, also our resources are quite tight as we have lots of projects using the same infrastructure. So we need to be mindful of others and any effects on productivity.

Secondly releasing code that is only meant for dev should be avoided, wherever possible.

In the past some of our good intentioned practices, have ended up making build stages lengthy and problematic at best.

andrewmee:
@jpw @Alicole I don't believe that having longer build times is a significant issue.  If we're adding something into our production code, it's highly likely to be a very small amount of code and so will be installed from npm in literally milliseconds.

If we were including something that's large enough to make any appreciable difference to build time, then it would be a much much larger problem to the performance of that app.

benjclark:
I'm in support of this proposal. I have long found dep and devDeps confusing and to have a documented clear boundary between the two will be of benefit to everyone. With this approach, it is easy to distinguish where a package should live and it will certainly lead to less confusion.

I agree with Andrew that the build time implications of this are likely to be negligible and do not outweigh the benefits of providing clarity on this across our organisation. I am always in favour of assisting the human rather than the machine. If we reduce ambiguity and introduce clarity we will become more productive.

Plus, as mentioned in the text, this is likely to become the way packages are differentiated when HTTP2/ES6/import is widely adopted and the norm. So let's get in the mindset.

jpw:
@andrewmee I tend to agree, and it smells of premature optimisation.

>In the past some of our good intentioned practices, have ended up making build stages lengthy and problematic at best.

@Alicole interesting point that we should be aware of, do you have any examples please?

Alicole:
@jpw -- over the years at BMC we went through various iterations of our build and deploy process.

If you have spent a whole day trying to get out what should be a simple build, you would know what I mean and then have to rope more and more resource to achieve it.

Also eventually one of our staging environments became littered with failed and aborted build attempts, that to deploy what should have been a simple change became like treading over a mine field.

jpw:
@Alicole OK thanks, that sounds terrible :) Maybe a conflation of performance and general flakiness by the sounds of it...  However as with all performance issues I think it's better to measure where the problems are and act appropriately, rather than prematurely optimise. That's best done by EE (although we shouldn't be doing things to knowingly make things terrible, obviously!), and node builds are a lot lighter than your average bag of JVM-based fun.

The counter-arg is that we are jeopardising prod with unwarranted version hikes and causing more developer confusion, which IMO is a stronger argument.

hollsk:
I would like to direct everyone for a brief moment to section 1, subsection b of the SN JavaScript House Style which is of relevance to this discussion 😉 

https://github.com/springernature/frontend-playbook/blob/master/javascript/house-style.md#code-for-humans

PR #211, opened 2018-01-24
Update SpeedCurve website link

kunalnagar:
SpeedCurve link is broken


jpw:
So it is, thanks very much for the contribution @kunalnagar :)

Issue #188, opened 2017-07-31
I was serious when I mentioned discussing "frontend" & "front-end"


hollsk:
I'm not convinced that there is a worthwhile reason for using two different formats to refer to things that are front-end related. 

**Pros and cons of using 2 different formats in particular circumstances:**

Pros:
* erm

Cons:
* People have to stop and think about which one to use
* People have to stop and think about whether the one used in a PR is correct or not
* It's something that needs to be _looked up_

--------------------------------------------------------

I don't really know how we got into this situation in the first place. Was it my fault? I hope it wasn't my fault. 

**Proposal:**

Use "frontend" for everything. I'll update the Playbook. Nobody speaks  or thinks about it ever again.

Pros:
* Everyone knows what to use
* It's still grammatically correct
* We will never need to speak of this or think about it ever again

Cons:
* People might not immediately find the content they're looking for if they search for "front-end" ¯\_(ツ)_/¯

hollsk:
#96  (hollsk opened this issue on 6 Oct 2016)

😂

andrewmee:
I don't think this is a problem

PR #213, opened 2018-02-27
house-stylify writing guide, break inclusive language into own file

hollsk:
Our head of Diversity and Inclusion has got me VERY excited.  This PR:

* breaks out section on writing inclusively into its own file (inclusive-language.md)
* changes "language.md" to "house-style.md" to bring it into line with the rest of the Playbook 
* couple of small text changes 
* updates references to the touched files

### DIFF
--- a/writing/house-style.md
+++ b/writing/house-style.md
@@ -14,7 +14,7 @@ We don't enforce strict standards on writing style as we want you to use your ow
 
 ### Tone of voice
 
-For a general audience, you should write like a human. Be professional, but conversational - we want our contributors and readers to feel comfortable joining in at any time. The exception to this is in writing [technical content](#technical-writing) for a technical audience.
+For a general audience, you should write like a human. Be professional, but conversational - we want our contributors and readers to feel comfortable joining in at any time.
 
 Be thankful to contributors and those who log issues. They're giving up their time to make our playbook better. Make sure they know we're grateful to them.
 
@@ -32,7 +32,7 @@ Follow Plain English guidelines to help safeguard the accessibility of your writ
   - Use contractions (e.g. "can't", "it's").
   - Go straight to the point.
 - Be specific
-  - Use the active voice instead of the passive voice (e.g. "NPM will install the dependencies" instead of "the dependencies will be installed").
+  - Prefer the active voice to the passive voice<sup>[\[1\]][active-passive]</sup> (e.g. "NPM will install the dependencies" instead of "the dependencies will be installed").
   - Address the user as "you" where possible.
   - Avoid referring to "things" or "stuff".
   - Be direct. Avoid metaphor, simile, and slang.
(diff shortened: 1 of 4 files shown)


writing/house-style.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+### Technical writing
+
+When expressing technical concepts, you'll need to use technical language. This may be complex, contain jargon, or seem formal. That's ok! Technical writing isn't for a general audience. Your priorities must be clarity, brevity, and accuracy, not the emotional impact of your words. 
+
+All of the advice in the [plain English](#plain-english) section applies. Additionally:
+
+- Expand acronyms the first time you use them, and limit their use thereafter e.g. "Avoid the use of TLAs (Three Letter Acronyms)".
+- When defining technical requirements, you may use the keywords defined in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) for those requirement levels, e.g. "MUST", "SHOULD", "OPTIONAL".
```
dotcode: I know you've used "e.g.", but is it worth also finishing this sentence with ", etc.", just to make it a bit more obvious that "MUST, SHOULD, OPTIONAL" are not the only options? Or should we list all the options? Those are questions, I don't know the answer, your decision. Etc.

writing/house-style.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Frontend or front-end
+
+* Use `front-end` when you need an adjective (e.g. "the front-end team", "front-end principles").
```
dotcode: Should Front-End Team be capitalised, as it's a specific thing?

writing/house-style.md
```diff
@@ -0,0 +1,60 @@
...
+
+
+## Writing style
+
+We don't enforce strict standards on writing style as we want you to use your own words to express your own personality. That said, we have three general sets of guidelines - tone of voice, plain English, and technical writing.
+
+
+### Tone of voice
+
+For a general audience, you should write like a human. Be professional, but conversational - we want our contributors and readers to feel comfortable joining in at any time. The exception to this is in writing [technical content](#technical-writing) for a technical audience.
```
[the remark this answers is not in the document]
hollsk: No, I agree. The exception in technical writing is that it probably shouldn't be _conversational_, but the wording muddles the concepts. I'm not sure how to fix it, though. 

Maybe we can just drop the entire "the exception to this..." sentence? Technical concepts _can_ be expressed conversationally, after all.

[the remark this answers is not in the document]
dotcode: Yeah good idea, lose it!

writing/house-style.md
hollsk: A _certain contributor_ 👀 noted in #188 that "... arguably it doesn't need to be in there - the most important thing with this is how we use it when naming things like repos..." which I don't disagree with. 

and I disagree with it even less when i think about going through the playbook and recapitalising every instance of "front-end team" 😂

hollsk: Soz, the purpose of my comment is unclear, isn't it?

To clarify, I'm objecting to changing the capitalisation on the grounds that the entire sentence/practice is unnecessary. Prescriptive enforcement of the two different types of use is a complication that IMO we'd be better off without. 

I'd prefer to sack the first instruction off entirely and only be prescriptive when it comes to repo names/slack channels/code.

 ¯\_(ツ)_/¯

Alicole: When it's used as a noun i.e. proper noun then that makes sense as capitalised.

Otherwise in common usage or as an adjective (where you are just qualifying another common noun) then lower case is fine.

jpw:
As far as I can see from a cursory glance the capitalisation of f-e has not changed in this PR, if we really want to go there again then maybe do so in a different PR?

PR #217, opened 2018-03-13
Update CONTRIBUTING.md

ambrwlsn:
Took out extra characters at the end of the CONTRIBUTING.md file

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,4 +30,4 @@ If you work at Springer Nature, please remember to keep discussion inside the is
 
 A bot notifies the #frontend Slack room when work is merged, because changes to the playbook are relevant to all Front-End developers at Springer Nature.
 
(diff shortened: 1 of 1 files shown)


sonniesedge:
Hi @ambrwlsn! You've exposed us as being dreadfully old here, as that's an equally old and equally dreadful [UNIX joke](https://en.wikipedia.org/wiki/Backspace#%5EH). 

It makes me think that not many people in this modern day and age will get it! 😭 

@dotcode @hollsk What are you thoughts?

hollsk:
It is a very techie joke, as well as an old one. We do want the Playbook to be accessible as far as possible to people who aren't necessarily developers.

I reckon we can lose the dad joke without loss of meaning 🙂

ambrwlsn:
@dotcode that is an amazing joke 🔥 😀 I had thought it was there for a reason. Definitely should have Googled beforehand. Shame there is no option to add a tooltip to the ^H^H^H^H^H^H^H^H with an explanation of the joke xD

Btw this is my first contribution to open source. Don't know whether it's an embarrassing start or not. Thanks everyone for being so nice about it 😄

PR #231, opened 2018-05-17
Remove patterns and redefine components

alexkilgour:
This updates the playbook to remove `patterns` from the way we write CSS. This was after discussion with the various teams using this style of css, and the potential confusion that arises from different people interpreting what is a `pattern` and what is a `component`.

Now we have only `components` and `utilities` which will hopefully be cleaner.

I have also updated the section on 'Combining components' to emphasise the need for avoiding coupling of components. We have had various discussion about how inheritance is going to work at a brand level and the need to avoid different packages depending on different versions of components.

Teams will still be able to use `mixins` at a product level to construct components, but we want to emphasise their self contained nature.

We can expand on this point when we write up guidelines for the different levels and how NPM inheritance is going to work.

### DIFF
--- a/css/how-we-write-css.md
+++ b/css/how-we-write-css.md
@@ -33,7 +33,7 @@ A Base rule is applied to an element using an element/type selector, a descendan
 #### Components
 Components use classes to map to specific UI elements.
 
-They should be **reusable** and independent of context i.e. they should not exist only on a specific part of the DOM tree or require the use of certain element types.
+They should be **reusable** and independent of context.
 
 They should be **modular**, which means they should have a single focus, and contain everything neccessary to display that part of the UI.
 
--- a/css/how-we-write-css.md
+++ b/css/how-we-write-css.md
@@ -88,6 +88,7 @@ We implement the following rules within each level:
 * Can include _private_ functions and variables (exist at this level)
 * Can include _private_ mixins (exist at _this_ level), not shared in the `30 Mixins` level
 * Can include child elements
+* MUST NOT include any associated mixins at the `30 Mixins` level
 
 #### 60 Utilities
 * MUST NOT include type selectors
--- a/css/how-we-write-css.md
+++ b/css/how-we-write-css.md
@@ -25,19 +25,22 @@ Contains all **global** SASS variables for your project e.g. colors, fonts, medi
 Contains all **global** SASS functions e.g. em calculation, unit stripping.
 
 #### Mixins
-Contains all **global** SASS mixins, this includes mixin definitions used in the **pattern** and **utility** levels.
+Contains all **global** SASS mixins, this includes mixin definitions used in the **component** and **utility** levels.
 
 #### Base
 A Base rule is applied to an element using an element/type selector, a descendant selector, or a child selector, along with any pseudo-classes. Base styles are related to the basic styles of your product, like typography, reset and global elements like links. Use this level to include any (third-party) resets or normalization css.
 
 #### Components
-Components use classes to map to specific UI elements. They should exist as stand-alone UI components and **never** depend on other components.
+Components use classes to map to specific UI elements.
 
-#### Patterns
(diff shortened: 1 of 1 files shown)


css/how-we-write-css.md
```diff
@@ -25,19 +25,22 @@ Contains all **global** SASS variables for your project e.g. colors, fonts, medi
...
 #### Base
 A Base rule is applied to an element using an element/type selector, a descendant selector, or a child selector, along with any pseudo-classes. Base styles are related to the basic styles of your product, like typography, reset and global elements like links. Use this level to include any (third-party) resets or normalization css.
 
 #### Components
-Components use classes to map to specific UI elements. They should exist as stand-alone UI components and **never** depend on other components.
+Components use classes to map to specific UI elements.
 
-#### Patterns
-Patterns use classes to apply a reusable design pattern. This level should include any pattern that does not bind itself to a particular UI element. Styles written here can include both structural and cosmetic rules, and apply branding.
+They should be **reusable** and independent of context.
```
sonniesedge: I'm not sure I 100% agree with this. e.g. a header component, or a main nav. Technically reusable, yes. But practically?

css/how-we-write-css.md
```diff
@@ -25,19 +25,22 @@ Contains all **global** SASS variables for your project e.g. colors, fonts, medi
...
+
+They should be **modular**, which means they should have a single focus, and contain everything neccessary to display that part of the UI.
```
danyalaytekin: 👋 necessary

css/how-we-write-css.md
```diff
@@ -128,40 +122,30 @@ HTML classes that have an impact only at specific screen sizes have a @breakpoin
 <p class="u-text-center u-text-left@small">Center align until small breakpoint, then align left.</p>
 ```
 
-## Patterns vs Utilties vs Components
+## Components vs Utilties
 
-The key difference between patterns and utilties are that utilites should be purely structural and contain no child elements, whereas patterns can combine both structural and cosmetic rules, and can make use of child and descendent selectors.
+The key differences between components and utilties are that utilites should be purely structural and contain no child elements, whereas components can combine both structural and cosmetic rules, and can make use of child and descendent selectors.
```
danyalaytekin: utilities * 2 👋

css/how-we-write-css.md
```diff
@@ -25,19 +25,22 @@ Contains all **global** SASS variables for your project e.g. colors, fonts, medi
...
-Patterns use classes to apply a reusable design pattern. This level should include any pattern that does not bind itself to a particular UI element. Styles written here can include both structural and cosmetic rules, and apply branding.
+They should be **reusable** and independent of context.
+
+They should be **modular**, which means they should have a single focus, and contain everything neccessary to display that part of the UI.
+
+They should be **isolated**, meaning they do not directly modify or depend on another component. This is more important than code reuse across components as this _could_ increase dependencies and tight coupling.
 
 #### Utilities
-Utilities are high-specificity, very explicit classes that apply a single rule or a single piece of functionality. They are used as overrides, helper classes, and to quickly build up simple page elements.
+Utilities are high-specificity, very explicit immutable classes that apply a single rule or a single piece of functionality. They are used as overrides, helper classes, and to quickly build up simple page elements.
```
sonniesedge: Maybe "quickly *scaffold* up simple page elements", to emphasise that building in this way should be more temporary and assists in building more solid components? 

Do we even want to come down on a side in the component vs atomic argument? (because our utility classes can easily be used as atomic units and I have View on this. I think using scoped components is 👍).

jpw: I think we should avoid the use of the word "atomic" due to connotations with other CSS approaches.

alexkilgour: I think using the classes from utilities has significant use cases, but i think the way this is structured _should_ avoid atomic uses

PR #234, opened 2018-05-23
Modules

o-t-w:
- Common practice to use the syntax `export {functionname, someotherfunctionname}` for named exports and `export default functionname` for default exports. Changed markdown text to reflect that. 
- You can only export one thing as default so `export default { get, isValid };` is not valid and has been changed to named exports.


alexkilgour:
Could you provide a description that details what changes are being made and why, thanks

zanonnicola:
Do you think is worth mention that if your module has more than one public function you should use `multiple named exports` ?

As I mention in the chat it allows tool like Rollup, Webpack to remove dead-code (aka Tree shaking).

Also, we probably should transpile the code already so we can ship the `umd` bundle to NPM. I'm not sure what's the policy for the FE component but usually you create different module format and add them to the package.json

```
module:  path to with ES6 modules
main:  path to CJS build
```

Don't think AMD is used anymore... 🤔

Doing this will allow your users to choose the most effective way to include your module based on their tool chain.

o-t-w:
Thanks @zanonnicola Can you approve the request? 😄

PR #237, opened 2018-05-30
adds shining gem of a statistic

hollsk:
I was noodling about in our intranet and came across a wonderful document posted by some people in charge of a product I've never heard of. They went for accessibility training and learned lots of fun things, including this very punchy statistic from Forrester Research that naturally I pilfered straight away. 

I've used it as a replacement for the link to FT Laura's talk as that's already linked to further down in the document.


dotcode:
Ooh, peer-reviewed source. Nice find.

hollsk:
It's a little bit cheaty as the pubmed source isn't the report itself - it just cites it. But it's easier to access the information there as it gives the relevant statistical summary and you don't need to jump through any hoops to access it. 

The report itself can be downloaded from this page: https://www.microsoft.com/en-us/download/details.aspx?id=18446 and it's a very worthwhile read. They've made an effort to include people who wouldn't see themselves as disabled but who do have some kind of impairment and who would benefit from AT and accessibility efforts.

PR #240, opened 2018-06-20
Add editorconfig with common settings for Markdown

josebolos:
So we can be consistent with the rest of the world

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false


joeyciechanowicz:
Doesn't the `@springernature/eslint-config` use tabs instead of spaces? Should we set that as the default then?

josebolos:
(For the record, I've just rewritten the file to make it clearer that we only care about Markdown files in this case.)

You are right, we do prefer tabs over spaces everywhere (well, at least for code). The only exceptions that we seem to be making are `package.json` files, as most tools (npm, snyk, etc.) expect the indentation to be 2 spaces, and will reformat the `package.json` file if you give them the chance. This is mentioned at the bottom of the [Linting section](https://github.com/springernature/frontend-playbook/blob/9a696591da665f475fab2ad9df48e0675c96f9b7/javascript/house-style.md#indentation) although it probably could be made clearer.

For Markdown files, we don't have any written down agreement, but the most popular choice seems to be 2 spaces, as it avoids accidentally indenting nested lists as code formatting or viceversa. We do tend to use this setting for README files in [some of our projects](
https://github.com/springernature/shunter/blob/master/.editorconfig), although we've been quite inconsistent with it. For example, most of the docs in this repo are currently using 4 spaces, with some using 2 spaces already, and a few using tabs.

This adds the `.editorconfig` so we can (eventually, with time) make the styling more consistent.

.editorconfig
```diff
@@ -0,0 +1,12 @@
+# http://editorconfig.org
+root = true
+
+[*]
+indent_style = space
```
jpw: sorry but how does this not apply spaces instead of tabs to all files?

i found this somewhere, seems to be closer, maybe start with this? (unless I have missed the point of this PR)

```
[*]
end_of_line = lf
charset = utf-8
indent_style = tab
indent_size = 4
insert_final_newline = true
trim_trailing_whitespace = true

[*.json]
insert_final_newline = false

[package.json]
indent_style = space
indent_size = 2

[*.md]
indent_style = space
trim_trailing_whitespace = false

[*.yml]
indent_style = space
indent_size = 2
```

josebolos: Yes, I created that file :) That was the original one that I was going to include here, but there are no files in this repo besides markdown files (and the LICENSE file), so I thought that it doesn't make much sense to include config for other files.

That said, maybe I should have included only the config for Markdown files. Or, it it makes it clearer, I can include the whole file.

jpw: lulz :) One file to rule them all, then link out to it?

josebolos:
Full file now included.

PR #244, opened 2018-06-22
Add RFC 2119

hollsk:
Been meaning to add this for quite some time

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Springer Nature front-end playbook
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this playbook are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
 ## What's this?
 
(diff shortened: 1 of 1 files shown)


dotcode:
One thought … have you trawled every page to ensure they are being used appropriately?

hollsk:
Nah. [RFC 8174](https://tools.ietf.org/html/rfc8174) states that RFC 2119 applies to only those words that are capitalised. 

The safest approach for us would be to allow that any key word that _isn't_ capitalised to have no special meaning, and to only capitalise in those instances where it does.

dotcode:
Wow, didn't know about RFC 8174. My opinion before we've had an in-depth debate where you prove me wrong: I think I dislike it. I've had many a moment in the past where I think "oh God, RFC 2119 applies, now I really have to think about this", to good effect. I also dislike the ambiguity this introduces. I'd rather just RFC 2119 applies, forcing people to properly consider their recommendations. 

How are you going to warp my fragile little mind?

hollsk:
I'm not. I just suggested it to try to get out of trawling the entire Playbook looking for keywords  💁‍♂️

jpw:
My 2p is I dislike it in non-formal documents because it imparts a VERY SHOUTY TONE.

danyalaytekin:
Maybe add a small apology above the new explanatory section, to emphasise the use of capitals is to comply with a convention rather than a result of rage?

jpw:
@danyalaytekin oooh I like that idea a lot, thanks!

PR #253, opened 2018-08-16
Simplify document and add note about TLS support

josebolos:
Welcome to this extremely lengthy PR! The aim of this PR is two-fold:

* Simplify the language of the document, in order to make it easier to understand to people already not familiar with our browser support strategy.
* Adding details about TLS, as it affects the way that we support old browsers.

List of individual changes:

Section 1: Introduction
============

* Rename document from "Graded browser support" to "Browser support". The reason is that I imagine that people looking for this document looking for this document will be looking for our "Browser support" policy, and may not necessarily familiar with what "Graded" is or means. And we don't offer an "Ungraded browser support" anyway, so there's not much sense on having "Graded" as part of the title. This makes things easier to find.
* Add a description of the purpose of the document before the table of contents

Section 2: Graded browser support
=================================

* Remove "Nate Koechley" as it's not relevant for understanding the criteria and afaik he's not even at Yahoo. There's also no mention of him in the document that we link to.
* Clarify the meaning of the different grades, replacing "modern" with "capable", and clarifying what a grade means in practical terms with regards to the experience ("core experience" vs "full experience").
* Move the browser list to a separate section, so it's easier to share a link to it.
* Add introducting text right before the browser list.
* Change "Android browser (Chromium)" to "Android browser (Webview)" as the Webview-based Android browser has not always been based on Chromium, it has only been based on Chromium since Android 4.4.
* Remove footnotes at the end of the support table, as they are mostly irrelevant. The "Edge" footnote has been addressed by replacing "n/a" with "none" in the Grade-C column (what does "not available" even mean in this context?). The IE10 footnote is part of an implementation detail, so it has been moved/merged to the "Implementation" section (which already mentions IE10).

Section 3: Implementation
=========================

* Add a paragraph explaining what the "cutting the mustard" technique is.
* Add a link explaining what "cutting the mustard" means, as it's unclear for non-native speakers.
* Add a paragraph describing in simple terms how we implement feature detection (using CSS media queries), and how it differs from the BBC approach which is JS-based.
* Move the CSS media query immediately after the bit that says "we use a CSS media query", as it makes the text flow better. This leaves the clarifications for later.
* Rephrase the "we load JS only if we load the enhanced CSS" bit so the sequence is clearer and the paragraph leads naturally to the example.
* Create a "Caveats" subsection in the implementation section. This contains both the details about IE10, and the restrictions due to lack of support for TLS 1.2. This whole section has been written from scratch, so I'm not sure about the wording though. Suggestions welcome!

Overall
=======

* Make all links part of the text, instead of a separate thing at the end of the sentence.
* Merge all the sentences that were on separate lines, but part of the same paragraph.
* Reduces the "Automated readability score" for the whole document from 12 to 10. 📖 🎉

If you're still reading...

madeit

### DIFF
--- a/practices/graded-browser-support.md
+++ b/practices/graded-browser-support.md
@@ -24,16 +24,16 @@ Graded Browser Support works by classifying all browser versions in one of three
 
 This is the current list of browser versions and their corresponding grades:
 
-| Tables                    | Grade-A support                 | Grade-C support |
-| ------------------------- |:-------------------------------:| ---------------:|
-| Chrome                    | latest stable, latest stable -1 | < 29            |
-| Edge                      | latest stable, latest stable -1 | none            |
-| Firefox                   | latest stable, latest stable -1 | < 29            |
-| IE                        | 11                              | < 11            |
-| Opera                     | latest stable                   | < 16            |
-| Safari iOS                | latest stable, latest stable -1 | < 7             |
-| Safari MacOS              | latest stable, latest stable -1 | < 6.1           |
-| Android browser (Webview) | Latest stable                   | < 4.4           |
+| Tables          | Grade-A support                 | Grade-C support    |
+| --------------- |:-------------------------------:| ------------------:|
+| Chrome          | latest stable, latest stable -1 | < 29               |
+| Edge            | latest stable, latest stable -1 | none               |
+| Firefox         | latest stable, latest stable -1 | < 29               |
+| IE              | 11                              | < 11               |
+| Opera           | latest stable                   | < 16               |
+| Safari iOS      | latest stable, latest stable -1 | < 7                |
+| Safari MacOS    | latest stable, latest stable -1 | < 6.1              |
+| Android Webview | Latest stable                   | < 30 (Android 4.4) |
 
 Every unlisted browser version is Grade-X.
 
--- a/practices/graded-browser-support.md
+++ b/practices/graded-browser-support.md
@@ -61,7 +61,7 @@ We couple the loading of JavaScript to the loading of the enhanced CSS. We use t
 
 ```javascript
 (function() {
-  var linkEl = document.queryElementByID('core-stylesheet');
+  var linkEl = document.getElementByID('core-stylesheet');
   if (window.matchMedia && window.matchMedia(linkEl.media).matches) {
     var script = document.createElement('script');
     script.src = 'your-script.js';
--- a/practices/graded-browser-support.md
+++ b/practices/graded-browser-support.md
@@ -85,6 +85,6 @@ This means that even if we consider IE10 a Grade-C browser, we currently serve I
 
 #### Browsers without TLS 1.2 support
 
-As of June 2018, all our sites are served through HTTPS using the [TLS 1.2 cryptographic protocol](https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.2) or newer. This means that users of browsers that don't support TLS 1.2 (e.g. Safari on iOS 4) or that have support for TLS 1.2 disabled (e.g. Internet Explorer on Windows 7) may not be able to access our sites. We consider that this is required in order to keep our users secure.
+As of June 2018, all our sites are served through HTTPS using the [TLS 1.2 cryptographic protocol](https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.2) or newer. This means that users of browsers that don't support TLS 1.2 (e.g. Safari on iOS 4) will not be able to access our sites. Browsers that have support for TLS 1.2 not enabled by default (e.g. Internet Explorer on Windows 7) will not be able to access our sites, unless they change their default settings. We consider that this is required in order to keep our users secure.
 
 The way that we restrict the connection to our sites when not using TLS 1.2 doesn't impact the way that we design and build our sites and our commitment to an approach based on progressive enhancement techniques.
--- a/practices/graded-browser-support.md
+++ b/practices/graded-browser-support.md
@@ -1,67 +1,90 @@
-# Graded browser support
+# Browser support
 
-* [Our criteria for browser support](#our-criteria-for-browser-support)
-* [How we implement graded browser support](#how-we-implement-graded-browser-support)
+This page describes the way that we support different browsers and browser versions when building sites at Springer Nature.
 
+* [Our criteria for browser support](#Our-criteria-for-browser-support)
+* [Graded browser support list](#Graded-browser-support-list)
+* [Implementing graded browser support](#Implementing-graded-browser-support)
+  * [Implementation details](#Implementation-details)
+  * [Caveats](#Caveats)
+    * [Internet Explorer 10 support](#Internet-Explorer-10-support)
+    * [Browsers without TLS 1.2 support](#Browsers-without-TLS-1.2-support)
 
 ## Our criteria for browser support
 
-We follow the guidelines for graded browser support outlined by Nate Koechley at Yahoo:
-[Yahoo Grade Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support)
+We follow the guidelines for graded browser support outlined by Yahoo in their [Graded Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support).
 
-Our browser support in summary:
+The graded browser support works by classifying all browser versions in one of three different grades:
 
-- Grade-A are modern and/or popular browser versions that receive all styling and behaviour. Bugs are addressed with high priority.
-- Grade-C are old and/or uncommon browser versions which are known to break if receiving full styling and behaviour. We serve only minimum CSS and JavaScript (detailed below). Bugs are addressed with high priority.
-- Grade-X browsers exist in the grey area between Grade-A and Grade-C. We serve them the same code as Grade-A, but we do not test our products against these browser versions. We expect our sites to work in these browsers. Over time, analysis of errors and usage patterns may cause some browsers in Grade-X to be deliberately changed to either Grade-A or Grade-C, to allow for better support. 
+* Grade-A are capable or very popular browser versions. These browsers will receive the full experience, with all styling and behaviour. Bugs are addressed with high priority.
+* Grade-C are incapable and/or uncommon browser versions which are known to break if receiving full styling and behaviour. These browsers will receive a core experience, with only minimum CSS and JavaScript. Bugs are addressed with high priority.
+* Grade-X browsers exist in the grey area between Grade-A and Grade-C. We serve them the same code as Grade-A, so they will receive the full experience, but we do not test our products against these browser versions. We expect our sites to work in these browsers. Over time, analysis of errors and usage patterns may cause some browsers in Grade-X to be deliberately changed to either Grade-A or Grade-C, to allow for better support.
 
-| Tables                      | Grade-A support                 | Grade-C support    |
-| --------------------------- |:-------------------------------:| ------------------:|
-| Chrome                      | latest stable, latest stable -1 | < 29               |
-| Edge                        | latest stable, latest stable -1 | n/a                |
-| Firefox                     | latest stable, latest stable -1 | < 29               |
-| IE                          | 11                              | < 11               |
-| Opera                       | latest stable                   | < 16               |
-| Safari iOS                  | latest stable, latest stable -1 | < 7                |
-| Safari MacOS                | latest stable, latest stable -1 | < 6.1              |
-| Android browser (Chromium)  | Latest stable                   | < 4.4              |
+## Graded browser support list
 
-### Notes
+This is the current list of browser versions and their corresponding grades:
 
-- We have no explicit Grade-C support for Edge.
-- We have no Grade-X support for IE, as we officially only support IE11, with anything below that being Grade-C. However, due to browser technical limitations we can only detect IE10+, so IE10 is, for now, included in our CTM media query. 
+| Tables                    | Grade-A support                 | Grade-C support |
+| ------------------------- |:-------------------------------:| ---------------:|
+| Chrome                    | latest stable, latest stable -1 | < 29            |
+| Edge                      | latest stable, latest stable -1 | none            |
+| Firefox                   | latest stable, latest stable -1 | < 29            |
+| IE                        | 11                              | < 11            |
+| Opera                     | latest stable                   | < 16            |
+| Safari iOS                | latest stable, latest stable -1 | < 7             |
+| Safari MacOS              | latest stable, latest stable -1 | < 6.1           |
+| Android browser (Webview) | Latest stable                   | < 4.4           |
 
-## How we implement graded browser support
+Every unlisted browser version is Grade-X.
 
-Our primary approach to graded browser support is the "cutting the mustard" progressive enhancement technique, based upon the principles developed by the BBC: [Cutting the Mustard](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard).  However, user agent sniffing is still applied on some legacy projects.
+## Implementing graded browser support
 
-We load a basic stylesheet to all users. This contains only [normalisation](https://necolas.github.io/normalize.css/) and a few small enhancements to the user-agent stylesheet.
-To load the full experience only in modern browsers (grade-A and grade-X) we implement logic in the media attribute of the `<link>` element that identifies the main stylesheet, loading the stylesheet only in browsers that recognise the properties of that media query.
-This technique is documented here: [Cutting the Mustard with Media queries](https://www.sitepoint.com/cutting-the-mustard-with-css-media-queries/).
+Our primary approach to graded browser support is the "[Cutting the Mustard (CTM)](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard)" progressive enhancement technique, based upon the principles developed by the BBC.
 
-The media queries we use are based upon [CSS Only Mustard Cut](https://github.com/Fall-Back/CSS-Mustard-Cut), with a preference towards combining them into one rather than separating them out into multiple `<link>` elements.  Note that you **cannot** add line breaks to the media query if they are combined.
+This approach works by using feature detection in order to determine which browsers (Grade-A and Grade-X) will receive the full experience (i.e. they "[cut the mustard](https://en.wiktionary.org/wiki/cut_the_mustard)") and which ones (Grade-C) will receive the core experience.
 
-Note: As mentioned above, while our graded browser support table specifies Internet Explorer 10 and below as Grade-C, until an updated media query is found this approach will render IE10 as Grade-A.
+### Implementation details
+
+All browser versions load a basic stylesheet that contains only [normalisation](https://necolas.github.io/normalize.css/) and a few small enhancements to the user-agent stylesheet.
+
+We then use CSS media queries to detect capable browsers. Unlike the approach described by the BBC, we don't use a JavaScript-based featured detection. We want users to have the best possible experience even if JavaScript is not available.
+
+To load the full experience only in grade-A and grade-X browsers we implement logic in the media attribute of the `<link>` element that identifies the main stylesheet, loading the stylesheet only in browsers that recognise the properties of that media query:
 
 ```html
 <link rel="stylesheet" href="your-css.css" media="only screen and (-webkit-min-device-pixel-ratio:0) and (min-color-index:0), (-ms-high-contrast: none), only all and (min--moz-device-pixel-ratio:0) and (min-resolution: 3e1dpcm)">
 ```
 
-We couple the loading of JavaScript to the loading of the enhanced CSS.
-If the browser loads the CSS within the media query, then load the JavaScript.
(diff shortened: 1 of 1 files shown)


practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
 
-- Grade-A are modern and/or popular browser versions that receive all styling and behaviour. Bugs are addressed with high priority.
-- Grade-C are old and/or uncommon browser versions which are known to break if receiving full styling and behaviour. We serve only minimum CSS and JavaScript (detailed below). Bugs are addressed with high priority.
-- Grade-X browsers exist in the grey area between Grade-A and Grade-C. We serve them the same code as Grade-A, but we do not test our products against these browser versions. We expect our sites to work in these browsers. Over time, analysis of errors and usage patterns may cause some browsers in Grade-X to be deliberately changed to either Grade-A or Grade-C, to allow for better support. 
+* Grade-A are capable or very popular browser versions. These browsers will receive the full experience, with all styling and behaviour. Bugs are addressed with high priority.
+* Grade-C are incapable and/or uncommon browser versions which are known to break if receiving full styling and behaviour. These browsers will receive a core experience, with only minimum CSS and JavaScript. Bugs are addressed with high priority.
```
sonniesedge: Maybe add something about their relation to web standards "These incapable and/or uncommon browser versions do not support the latest web standards and are known to break if receiving full styling and behaviour.

Do we address Grade-C bugs with high priority? Does it perhaps mean "Bugs in the core experience are addressed with high priority"?

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
 ```javascript
 (function() {
-    var linkEl = document.querySelector('link');
-    if (window.matchMedia && window.matchMedia(linkEl.media).matches) {
-        var script = document.createElement('script');
-        script.src = 'your-script.js';
-        script.async = true;
-        document.body.appendChild(script);
-    }
+  var linkEl = document.querySelector('link');
```
sonniesedge: I think we should take the chance to change this so that the appropriate <link> is specified by an ID attribute. I've seen people try to implement this when there are multiple <link>s on a page, and things have broken as a result.

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
-Note: As mentioned above, while our graded browser support table specifies Internet Explorer 10 and below as Grade-C, until an updated media query is found this approach will render IE10 as Grade-A.
+### Implementation details
+
+All browser versions load a basic stylesheet that contains only [normalisation](https://necolas.github.io/normalize.css/) and a few small enhancements to the user-agent stylesheet.
+
+We then use CSS media queries to detect capable browsers. Unlike the approach described by the BBC, we don't use a JavaScript-based featured detection. We want users to have the best possible experience even if JavaScript is not available.
+
+To load the full experience only in grade-A and grade-X browsers we implement logic in the media attribute of the `<link>` element that identifies the main stylesheet, loading the stylesheet only in browsers that recognise the properties of that media query:
 
 ```html
```
sonniesedge: See next comment:

```html
<link rel="stylesheet" href="your-css.css" media="only screen and (-webkit-min-device-pixel-ratio:0) and (min-color-index:0), (-ms-high-contrast: none), only all and (min--moz-device-pixel-ratio:0) and (min-resolution: 3e1dpcm)" id="corestylesheet">
```

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
 ```
 
-The javascript code outlined above will fail in Internet Explorer if there is more than one link element. You can get around this by targeting the specific link element using a class or ID.
+Please note that this JavaScript code will fail in Internet Explorer if there is more than one link element. You can get around this by targeting the specific link element using a class or ID.
+
+### Caveats
+
+#### Internet Explorer 10 support
+
+As on January 2017, we consider Internet Explorer 10 as grade-C. This is mainly due to its lack of support for certain modern features plus a very low usage. IE10 also stopped receiving security updates in January 2016 from its manufacturer.
```
hollsk: Typo "As on January 2017" should be "As _of_..."

hollsk:
vg what an epic pr 😍

practices/graded-browser-support.md
josebolos: > Maybe add something about their relation to web standards "These incapable and/or uncommon browser versions do not support the latest web standards and are known to break if receiving full styling and behaviour.

Ooh, I like that. Although maybe not say "are known to break" as usually when we classify a browser as Grade C is because someone wanted to do or use something specific, and that browser version breaks when using that thing. That said, I'm not sure what to replace it with.

> Do we address Grade-C bugs with high priority? Does it perhaps mean "Bugs in the core experience are addressed with high priority"?

Ah, this is one of these things that has been copied and pasted from document to document to document and it has lost its meaning. My understanding is that:

* Grade A: We test on them, we fix bugs that affect them with high priority.
* Grade C: We test on them, we fix bugs that affect them with high priority.
* Grade X: We don't test on them, we fix bugs that affect them with normal priority.

@dotcode Do you mind confirming if my understanding is correct? ^

We could probably explain that in a slightly better way.

practices/graded-browser-support.md
josebolos: 🤷 Bear in mind I don't use any of this stuff, if you provide me with the right code I can just slap it there.

practices/graded-browser-support.md
josebolos: 🤷 Same as above, if that's the correct code I will just slap it there.

dotcode:
Very nice Jose, thanks :) Two questions unrelated to the changes:

1. once you get this approved please may you also edit the related page on our beautiful intranet?
2. how do you test the "automated readability" score?

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+  * [Implementation details](#Implementation-details)
+  * [Caveats](#Caveats)
+    * [Internet Explorer 10 support](#Internet-Explorer-10-support)
+    * [Browsers without TLS 1.2 support](#Browsers-without-TLS-1.2-support)
 
 ## Our criteria for browser support
 
-We follow the guidelines for graded browser support outlined by Nate Koechley at Yahoo:
-[Yahoo Grade Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support)
+We follow the guidelines for graded browser support outlined by Yahoo in their [Graded Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support).
```
jpw: Why not give Nate props?

jpw: Also it should be Graded Browser Support (capitalised) throughout this doc.

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+| Tables                    | Grade-A support                 | Grade-C support |
+| ------------------------- |:-------------------------------:| ---------------:|
+| Chrome                    | latest stable, latest stable -1 | < 29            |
+| Edge                      | latest stable, latest stable -1 | none            |
+| Firefox                   | latest stable, latest stable -1 | < 29            |
+| IE                        | 11                              | < 11            |
+| Opera                     | latest stable                   | < 16            |
+| Safari iOS                | latest stable, latest stable -1 | < 7             |
+| Safari MacOS              | latest stable, latest stable -1 | < 6.1           |
+| Android browser (Webview) | Latest stable                   | < 4.4           |
```
jpw: Not sure why this changed from Chromium to Webview? You have the browser bundled with the OS (Chrome) and the browser engine that powers webviews (a chromeless browser used by hybrid apps), which is now powered by Chromium.

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+* Grade-A are capable or very popular browser versions. These browsers will receive the full experience, with all styling and behaviour. Bugs are addressed with high priority.
+* Grade-C are incapable and/or uncommon browser versions which are known to break if receiving full styling and behaviour. These browsers will receive a core experience, with only minimum CSS and JavaScript. Bugs are addressed with high priority.
+* Grade-X browsers exist in the grey area between Grade-A and Grade-C. We serve them the same code as Grade-A, so they will receive the full experience, but we do not test our products against these browser versions. We expect our sites to work in these browsers. Over time, analysis of errors and usage patterns may cause some browsers in Grade-X to be deliberately changed to either Grade-A or Grade-C, to allow for better support.
```
jpw: I would change to;

> * Grade-X browsers are everything not Grade-A or Grade-C—such as bleeding-edge or very niche browsers. We serve them the same code as Grade-A, so they will receive the full experience, but we do not test our products against these browser versions. We expect our sites to work in these browsers. Over time, analysis of errors and usage patterns may cause some browsers in Grade-X to be deliberately changed to either Grade-A or Grade-C, to allow for better support.

practices/graded-browser-support.md
jpw: `core-stylesheet` ?

practices/graded-browser-support.md
jpw: `Grade-C`

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+    * [Browsers without TLS 1.2 support](#Browsers-without-TLS-1.2-support)
 
 ## Our criteria for browser support
 
-We follow the guidelines for graded browser support outlined by Nate Koechley at Yahoo:
-[Yahoo Grade Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support)
+We follow the guidelines for graded browser support outlined by Yahoo in their [Graded Browser Support guide](https://github.com/yui/yui3/wiki/Graded-Browser-Support).
 
-Our browser support in summary:
+The graded browser support works by classifying all browser versions in one of three different grades:
```
jpw: > Graded Browser Support works...

no "the"

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
 
-Our primary approach to graded browser support is the "cutting the mustard" progressive enhancement technique, based upon the principles developed by the BBC: [Cutting the Mustard](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard).  However, user agent sniffing is still applied on some legacy projects.
+## Implementing graded browser support
 
-We load a basic stylesheet to all users. This contains only [normalisation](https://necolas.github.io/normalize.css/) and a few small enhancements to the user-agent stylesheet.
-To load the full experience only in modern browsers (grade-A and grade-X) we implement logic in the media attribute of the `<link>` element that identifies the main stylesheet, loading the stylesheet only in browsers that recognise the properties of that media query.
-This technique is documented here: [Cutting the Mustard with Media queries](https://www.sitepoint.com/cutting-the-mustard-with-css-media-queries/).
+Our primary approach to graded browser support is the "[Cutting the Mustard (CTM)](http://responsivenews.co.uk/post/18948466399/cutting-the-mustard)" progressive enhancement technique, based upon the principles developed by the BBC.
```
jpw: capitalise GBS

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+This approach works by using feature detection in order to determine which browsers (Grade-A and Grade-X) will receive the full experience (i.e. they "[cut the mustard](https://en.wiktionary.org/wiki/cut_the_mustard)") and which ones (Grade-C) will receive the core experience.
 
-Note: As mentioned above, while our graded browser support table specifies Internet Explorer 10 and below as Grade-C, until an updated media query is found this approach will render IE10 as Grade-A.
+### Implementation details
+
+All browser versions load a basic stylesheet that contains only [normalisation](https://necolas.github.io/normalize.css/) and a few small enhancements to the user-agent stylesheet.
+
+We then use CSS media queries to detect capable browsers. Unlike the approach described by the BBC, we don't use a JavaScript-based featured detection. We want users to have the best possible experience even if JavaScript is not available.
+
+To load the full experience only in grade-A and grade-X browsers we implement logic in the media attribute of the `<link>` element that identifies the main stylesheet, loading the stylesheet only in browsers that recognise the properties of that media query:
```
jpw: capitalise Grade throughout

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+
+As on January 2017, we consider Internet Explorer 10 as grade-C. This is mainly due to its lack of support for certain modern features plus a very low usage. IE10 also stopped receiving security updates in January 2016 from its manufacturer.
+
+Unfortunately, both IE10 and IE11 support the same CSS media queries, which means that it's not possible to use a CSS-only method of differentiating between both browsers.
+
+This means that even if we consider IE10 a Grade-C browser, we currently serve IE10 users the full experience.
+
+#### Browsers without TLS 1.2 support
+
+As of June 2018, all our sites are served through HTTPS using the [TLS 1.2 cryptographic protocol](https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.2) or newer. This means that users of browsers that don't support TLS 1.2 (e.g. Safari on iOS 4) or that have support for TLS 1.2 disabled (e.g. Internet Explorer on Windows 7) may not be able to access our sites. We consider that this is required in order to keep our users secure.
```
jpw: "that have support for TLS 1.2 disabled by default (e.g. Internet Explorer 10 on"

practices/graded-browser-support.md
sonniesedge: omg, I totes gave you the code in the comments! it's almost the same as what's there, but with IDs as the selectors, rather than generic link elements.

practices/graded-browser-support.md
```diff
@@ -1,67 +1,90 @@
...
+
```
jpw: Maybe add this?

> Note that for IE 8, 9 & 10 on Windows 7 & 8, users can enable TLS 1.2 support themselves: _Tools > Internet Options > Advanced > Settings (Security Section)_. 

It might help a user out there... somewhere...

practices/graded-browser-support.md
josebolos: Well, the Android browser has always been Webview, but Webview only started using Chromium as of Android 4.4. Older versions of Webview didn't use Chromium, and we specifically added that browser to the list as we made versions older than 4.4 (so, the non-Chromium based ones) Grade-C.

practices/graded-browser-support.md
josebolos: I thought about it for a long time while I was writing but I decided to omit "by default", as the sentence reads:

> [...] users of browsers [...] that have support for TLS 1.2 disabled may not be able to access our sites.

Which is technically correct, the best kind of correct. However, saying:

> [...] users of browsers [...] that have support for TLS 1.2 disabled _by default_ may not be able to access our sites.

Is not always true, as if the user has enabled support for TLS 1.2, they will be able to access the content just fine.

Meaning that in this context, the only thing that matters is if the user's browser is actually supporting TLS 1.2 or not. I think that the fact that it's disabled by default (or not) is not relevant.

(There are also other browser & OS combos that support TLS 1.2 but it's not enabled by default, so this is not specific about IE10.)

practices/graded-browser-support.md
josebolos: I don't feel like that's in the scope of this document, which is mostly browser support and its implementation details? I wouldn't expect users to find this document.

Additionally, there are other browser & O.S. combos that support TLS 1.2 but don't have the support enabled by default, so this is not specific about IE10. Shirley we don't want to add all the possible ways to enable TLS 1.2 in all the possible browsers?

practices/graded-browser-support.md
josebolos: You have a point, the sentence as it currently stands it's not great:

> Grade-X browsers exist in the grey area between Grade-A and Grade-C.

I'm not sure about wording it as "bleeding-edge or very niche browsers". Chrome 62, for example, is Grade-X, but I wouldn't consider it bleeding-edge or very niche.

I was discussing the wording of "modern" vs "capable" vs whatever with @dotcode yesterday, so maybe he has a better suggestion than me.

josebolos:
Implement some of the changes requested:

* Fix a typo
* Capitalise all instances of "Graded Browser Support"
* Capitalise all instances of "Grade-?"
* Update CSS media query and JS snippet to use an id

practices/graded-browser-support.md
jpw: The wording is from my memory on Grade-X when GBS was written 🤷‍♂️  The thing is the wording implies X is somewhere between A and C, which isn't really the case.

practices/graded-browser-support.md
jpw: > I don't feel like that's in the scope of this document, which is mostly browser support and its implementation details? I wouldn't expect users to find this document.

True, but if say customer care is contacted by a user there is some advice for them. The playbook is useful for a reference we can point people at, as well as a thing that can be discovered.

> Additionally, there are other browser & O.S. combos that support TLS 1.2 but don't have the support enabled by default, so this is not specific about IE10. Shirley we don't want to add all the possible ways to enable TLS 1.2 in all the possible browsers?

No, naturally, but IE users are going to be the biggest chunk of those users, and I don't think it does any harm to offer help for these poor souls.

practices/graded-browser-support.md
jpw: >>[...] users of browsers [...] that have support for TLS 1.2 disabled by default may not be able to access our sites.

>Is not always true

I think it's always true, because if it is disabled by default they _may not be able to access our sites, depending on their settings_. Possible confusion around "able to" meaning "actually can" versus "potentially can".

I think the issue is that sentence is trying to cater for two cases -- complete incompatibility, and possible compatibility. How about, instead of;

> This means that users of browsers that don't support TLS 1.2 (e.g. Safari on iOS 4) or that have support for TLS 1.2 disabled (e.g. Internet Explorer on Windows 7) may not be able to access our sites.

we try

> This means that users of browsers that don't support TLS 1.2 (e.g. Safari on iOS 4) will not be able to access our sites. Browsers that have support for TLS 1.2 not enabled by default (e.g. Internet Explorer on Windows 7) will not be able to access our sites, unless they change their default settings.

Then in the following text we give them a hint on how to do that 🎉

practices/graded-browser-support.md
josebolos: > True, but if say customer care is contacted by a user there is some advice for them. The playbook is useful for a reference we can point people at, as well as a thing that can be discovered.

Sure, though I think that advice for customer care would better belong to our intranet than here.

practices/graded-browser-support.md
jpw: Yep, but "android browser" and "webview" are different things, which is not what this table row implies... Chrome releases across platforms have parity, so I think that's encompassed by the "Chrome" entry at the top of the table. So this column row is pretty much redundant these days, apart from calling out old webkit-based webviews.

So these days I think this row is about webview, not the browser -- especially as that is what the version number in the Grade-C column is referring to. How about we change the text from "Android browser (Webview)" to just "Android Webview"? Also the version number in the final column is not a browser version, but an OS version, so [maybe change that to "Android  < 4.4.3"](https://developer.chrome.com/multidevice/webview/overview#webview_faq)

And yes, I did read your PR notes 😁

josebolos: I agree, the "Android browser" description is not right. I've replaced it with "Android Webview" now. I've also found the version, which is 30.0.0.0, so I've made it "< 30 (Android 4.4)".

Ref: https://developer.chrome.com/multidevice/webview/overview

josebolos:
Added "Change Android Browser to Android Webview"

josebolos:
Added back Nate.

hollsk:
Here's a working Wayback Machine link to Nate's actual blog post, which might make sense to include so people know why we're naming him in the document 😉 

https://web.archive.org/web/20060304042737/http://developer.yahoo.net/yui/articles/gbs/gbs.html

josebolos:
Added link to gbs on archive.org.

josebolos:
"Automated readability score" is now 11. :(

practices/graded-browser-support.md
josebolos: That's a good one, I've committed that now. As mentioned, I don't feel comfortable adding instructions for the users, but feel free to submit a PR so others can voice their opinion too!

josebolos:
I've just pushed a change that rewrites the TLS 1.2 support bit with @jpw's suggestion. If I'm not mistaken that means that all the concerns raised have now been addressed (apologies if I've missed anything!).

The only two things left are the following comments, which could benefit from @dotcode's input:
* https://github.com/springernature/frontend-playbook/pull/253#discussion_r210672842
* https://github.com/springernature/frontend-playbook/pull/253#discussion_r210855802

practices/graded-browser-support.md
dotcode: @josebolos your understanding is correct.

josebolos:
I've just fixed the CTM example as @o-t-w pointed out.

Could people please review this so it can be merged? Thanks!

josebolos:
press-the-button

practices/graded-browser-support.md
sonniesedge: I'm up for adding the IE10 TLS config here. It _could_ help the bulk of users with an issue, and acts as a public place to point customer care and customers.  

I guess it shows that we're aware that an exception to our "100% coverage of all users" philosophy is caused by an erroneous browser setting and not by us doing something "wrong".

PR #258, opened 2018-09-04
Fix JS function name

zanonnicola:
Just a small typo on the function name :)


dotcode:
Oh wow. I had to look at that for a long time before I saw it.

PR #268, opened 2018-10-24
Streamlining JavaScript house style

davidpauljunior:
### Summary

This PR attempts to do the following:
- Bring code examples up to date but using `const` and `let` instead of `var`
- Make use of ES2015 code the 'norm', and have exceptions added if products _don't_ support it
- Remove anything that is / should be caught by linting
- Remove anything that simply states that _you can use x ES feature_

### Additional notes

I made a change to the `obj.hasOwnProperty` rule because of a [side note mentioned by eslint](https://eslint.org/docs/rules/guard-for-in).
> Note that simply checking foo.hasOwnProperty(key) is likely to cause an error in some cases

Admittedly, that code snippet could be removed entirely from the document and just use linting to enforce the rule.  There is a bit of a blurry line between where documentation finishes and linting begins.  But this is a first pass at trying to improve that 😄.


javascript/house-style.md
```diff
@@ -343,51 +342,33 @@ We use automated tools (like Snyk) to monitor the status of dependencies in our
...
-
-var baz = function() {
-    // ...
-}
+const foo = 1,
+      bar = 2;
 ```
 
-### Variables
+We use `let` only when a variable _explicitly_ [needs to be mutable](https://ada.is/blog/2015/07/13/immutable/):
```
hollsk: Is there a missing example here? The colon on the end suggests so 😉

javascript/house-style.md
```diff
@@ -343,51 +342,33 @@ We use automated tools (like Snyk) to monitor the status of dependencies in our
...
+
+#### Use of var
 
-We define variables with multiple `var` statements and trust our developers to understand [hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting). We generally discourage defining variables inside loops or blocks.
+If the environment you're running in does not support ES2015 (Node.js 4.x/[Babel](https://babeljs.io/)), and therefore you are using `var` instead of `const` and `let`, we trust our developers to understand [hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting).
```
hollsk: This sentence is a bit tortured. I know it's not your original work though, so you are blameless 😄 

```suggestion
If you have to use `var` instead of `const` and `let` (because your environment doesn't support ES2015 (Node.js 4.x/[Babel](https://babeljs.io/))), you need to understand [hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting). 
```

javascript/house-style.md
```diff
@@ -343,51 +342,33 @@ We use automated tools (like Snyk) to monitor the status of dependencies in our
...
 ```
 
-### Variables
+We use `let` only when a variable _explicitly_ [needs to be mutable](https://ada.is/blog/2015/07/13/immutable/):
+
+#### Use of var
 
-We define variables with multiple `var` statements and trust our developers to understand [hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting). We generally discourage defining variables inside loops or blocks.
+If the environment you're running in does not support ES2015 (Node.js 4.x/[Babel](https://babeljs.io/)), and therefore you are using `var` instead of `const` and `let`, we trust our developers to understand [hoisting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/var#var_hoisting). 
+We generally discourage defining variables inside loops or blocks.
```
hollsk:
```suggestion
Avoid defining variables inside loops or blocks.
```

(makes the instruction stronger ✊)

javascript/house-style.md
davidpauljunior: Ah yes, I found the example given a bit meaningless.

```js	
let foo = 'bar';	
```

davidpauljunior: Maybe this is an ok example?

We do this:

```js
let foo = 1;

if (foo) {	
    foo += 1;	
}
```

```js
const foo = [];

if (bar) {
    foo.push(1);
}
```

We _don't_ do this:

```js
let foo = [];

if (bar) {
    foo.push(1);
}
```

hollsk: maybe using a loop would be a good way to use a more meaningful example? e.g.

```
for(let i = 0; i < 10; i++)  { }
```

I'm not wedded to this suggestion, though - I'm only making it because it would make sense to have an example if `const` and `let` had their own headings! But they don't _need_ separate headings, especially if only one has a meaningful code example 🤷‍♀️

hollsk: oh you beat me to the comment 😄 

Yeah, i like those examples!

davidpauljunior: I actually changed the heading to "Defining variables".
What do you think? https://github.com/springernature/frontend-playbook/pull/268/commits/356330ab79cfc15e0d0e3ba5d282c4111643768f

hollsk: yeah, nicely done 👍

PR #275, opened 2018-12-19
Be more welcoming

dotcode:
Given this is now the top Google result for "frontend playbook", I think less prominent mention of ownership is appropriate. It now could be seen as the discipline's de-facto standard frontend playbook.

Advertising/treating it as such, may be beneficial to us (as it will present a more welcoming face), and to others (as they may feel more comfortable utilising it if they can refer to it as a standard, rather than simply a company's preferred way of doing things - which it is as well, but … you know, it's also utilised by many others).

Lots of justification for deleting two words, but that's how I work innit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# Springer Nature front-end playbook
+# The Front-End Playbook
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this playbook are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 
 ## What's this?
 
-This repo contains the Springer Nature front-end playbook. It details how we run software development and how we make web and mobile products together. It's filled with things we've learned based on our own experience and study of others' experiences.
+This repo contains The Front-End Playbook. It details how we run software development and how we make web and mobile products together. It's filled with things we've learned based on our own experience and study of others' experiences.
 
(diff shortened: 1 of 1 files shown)


hollsk:
>It now could be seen as the discipline's de-facto standard frontend playbook.

If this is the case then we'd need to comb the Playbook and excise all mentions of our org-specific practices, no? 

E.g. https://github.com/springernature/frontend-playbook/blob/master/CONTRIBUTING.md

>If you work at Springer Nature, please remember to keep discussion inside the issues and pull requests, avoiding Slack, hallway conversations etc. Remember that this repo is public, and the discussions we have can be of benefit to people apart from us.

>A bot notifies the #frontend Slack room when work is merged, because changes to the playbook are relevant to all Front-End developers at Springer Nature.

etc

sonniesedge:
Setting us up as _the_ documentation for frontend development on the planet. Let no one ever accuse you of thinking small, @dotcode. 😂 

If one of the criteria is "more welcoming" then updating the intro might be an easy win as well?
image

Because I would run if I was a first-time viewer.

sonniesedge:
> If this is the case then we'd need to comb the Playbook and excise all mentions of our org-specific practices, no?

😱

dotcode:
> we'd need to comb the Playbook and excise all mentions of our org-specific practices, no?

No … I'm not suggesting removing _all_ our mentions of org-specific practices. The goal is just to drop the prominence a bit.

dotcode:
Charlie you're plainly evil for suggesting we lose the RFC 2119 reference, but … I guess you're right. Still need it mentioned somewhere, but maybe at the bottom rather than the top? Open a PR.

dotcode:
> Setting us up as the documentation for frontend development on the planet. Let no one ever accuse you of thinking small, @dotcode. 😂

PR #277, opened 2019-01-09
Updates the CSS linting

davidpauljunior:
Similar to the [removal of the JS documentation](https://github.com/springernature/frontend-playbook/pull/268/files) that would be handled by linting rules, this removes the documentation about CSS linting rules.


css/house-style.md
```diff
@@ -97,6 +83,17 @@ We do this:
 
 Don't use `!important`. If you must, leave a comment, and prioritise resolving specificity issues before resorting to `!important`.
 
+### Linting
+
+Static analysis tools like linters can flag programming errors, bugs and stylistic errors, making your code more robust, readable and maintainable.
+
+CSS code (not compiled) should be linted with [styleLint](https://github.com/stylelint/stylelint). You can find _an example_ config file in the [config directory](config).
+
+SASS code should be linted with [sass-lint](https://github.com/sasstools/sass-lint) using the [Springer Nature
```
sonniesedge: Would the sass-lint linting work on plain CSS, I wonder? If it did we could use that and get rid of the reference to `styleLint`.

davidpauljunior: I don't think so.  I _think_ `stylelint` can work with both, but `sasslint` only with `sass` and `scss`.
I wanted to get rid of the stylelint part too.  I'm wondering whether it can go anyway.  This is aspirational and I don't believe we encourage people to write in CSS, do we?

davidpauljunior: But the play book is aspirational

davidpauljunior: Paging @alexkilgour, what do you think about the above comments? Specifically, can I get rid of the CSS/stylelint part?

sonniesedge: I  think having a separate linter for CSS to be a bit confusing. Like Joey says, I'm sure there are teams writing pure CSS. Maybe the aspiration bit is getting them onto Sass...?

davidpauljunior: Maybe this is too black and white, but the way I see it is one the following:

a) New projects can be authored in CSS, in which case we should have a `stylelint-config` repo instead of the config in the playbook.
b) New projects cannot be authored in CSS, in which case we shouldn't have the stylelint config in the playbook, but we could add a note saying that legacy code should be linted using something like stylelint.

joeyciechanowicz: How about:

c) New projects cannot be authored in CSS. Old projects that use CSS should still at least lint them using `stylelint`.

joeyciechanowicz: `sass-lint` also lints css just fine.

image

davidpauljunior: @joeyciechanowicz are the 2 extra errors on the `.scss` file specific to `scss`?  Like on a nested selector or something?

joeyciechanowicz: Interesting. Checked with exactly the same file and you get different counts of errors.

What is it that `scss-lint` offers that can't be done in `stylelint`? As theres the [stylelint-scss](https://github.com/kristerkari/stylelint-scss) plugin and [stylelint-config-recommended-scss](https://github.com/kristerkari/stylelint-config-recommended-scss) config etc

alexkilgour: > stylelint by itself supports SCSS syntax very well (as well as other preprocessors' syntaxes). Moreover, it introduces some specific rules that can be used to lint SCSS, e.g. to limit nesting, control the way @-rules are written. Yet stylelint is in general focused on standard CSS

So I _think_ that while stylelint supports some sass linting it is very limited

alexkilgour: I would think we could use `sass-lint` on css as well as sass. If you look at [our config](https://github.com/springernature/sasslint-config-springernature/blob/master/config/.sass-lint.yml) then a lot of those rules apply as much to css as much as to sass. We could run stylelint with plugins for scss, but it requires plugins, more dependencies to manage etc...

If we have a project already linted with `sass-lint` could be run the linter on the compiled css (non-minified) and see if it passes?

alexkilgour: It would be nice if we could settle on one though. I think historically stylelint was used (as we didn't use sass), then any new projects that started out from scratch with sass used sass-lint. I think I tried to use stylelint to replicate the sass-lint rules that were already in use but couldn't get it to work (that my be different now).

sonniesedge: @davidpauljunior What do you think? Looks like we can use `sass-lint` on the CSS as well and get rid of the CSS-specific section.

davidpauljunior: @sonniesedge I spoke to @alexkilgour about this earlier.
We'd need to test whether sass-lint works as we're expecting on CSS code. That could potentially be a rabbit hole. So for the sake of this PR, I will leave in the sentence about CSS and the stylelint config, then create an issue in the playbook to investigate the sass-lint part further.
It also feels a bit odd to me that projects which are using CSS would lint their code using something called 'sass-lint'.  Just doesn't feel right.

PR #297, opened 2019-04-05
Remove nesting section in BEM CSS part

### LINKED ISSUE #296: Remove nesting section in BEM CSS part of the Frontend Playbook
Link in the playbook: https://github.com/springernature/frontend-playbook/blob/master/css/bem-css.md#nesting

As discussed in the Frontend open space (https://springernature.slack.com/archives/C1HGRAQ8G/p1554382920003100) this section can be removed.

In a nutshell, reasons from the discussion, for that change to happen are:
- "we don’t use this practise anymore"
- It prevents from "being able to find the class names easily"
- "it hinders FEDs understanding the context that they are working on"
- "not-nesting names seems to have gained a lot of fans in fed in general lately."
- "When reviewing a PR, it requires extra effort to see the actual selector and to do a proper review" (e.g. https://pbs.twimg.com/media/D2xTdNLVAAAl3VG.png)

morgaan:
Close #296

### DIFF
--- a/css/bem-css.md
+++ b/css/bem-css.md
@@ -85,38 +85,3 @@ Choose your modifiers wisely. These two rules have very different meaning:
 .c-block--modifier .c-block__element { color: red; }
 .c-block__element--modifier { color: red; }
 ```
-
-## Nesting
-
-If you are using a CSS pre-processor, excessive nesting can be avoided by using smart class naming (with the help of BEM). Avoid nesting selectors more than _3 levels_ deep, and prefer using nesting as a convenience to extend the parent selector over targeting nested elements. For example:
-
-```scss
-.c-block {
-    padding: 24px;
-
-    &--modifier {
-        padding: 12px;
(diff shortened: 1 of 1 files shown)


davidpauljunior:
I think that instead of just deleting this section we may need to explicitly say that we _don't_ do nesting when using a preprocessor, and then list the reasons you raised in the issue.
And a block of code to demonstrate.

We do this:

```
.c-block { padding: 24px; }

.c-block--modifier { padding: 12px; }

.c-block__element { color: white; }

.c-block__element--modifier { color: black }
```

We _don't_ do this:

```
.c-block {
    padding: 24px;

    &--modifier {
        padding: 12px;
    }

    &__element {
        color: white;

        &--modifier {
            color: black;
        }
    }
}
```

morgaan:
Thank you @davidpauljunior for the feedback. Just committed the change according to what you said.

PR #298, opened 2019-04-05
Add example to strict mode usage with IIFE

morgaan:


javascript/house-style.md
```diff
@@ -538,6 +538,15 @@ In Node.js, you can add `'use strict';` to the top of each of your source files.
 
 In-browser you should not add your `'use strict';` to the top of the file; instead you should add it to either the file's [IIFE](http://benalman.com/news/2010/11/immediately-invoked-function-expression/) or each defined function if an IIFE isn't present. This is because global strict mode in browsers can cause issues with third-party code.
 
+We do this:
+
+```js
+(function () {
+  'use strict';
```
davidpauljunior: Tabbing is a bit off here (I think, although github can be odd about this things in diffs).
At least, the indentation of the function contents looks a bit different to other examples in the same file.

morgaan: Oops! Just fixed that. At least I guess

PR #308, opened 2019-06-20
Yes, you can use javascript

hollsk:
Not only that, but you _should_ use javascript! But that's not the same thing as assuming that everyone else _has_ javascript 👀


axemonkey:
Muy bien. 🎉

PR #311, opened 2019-06-27
Add document about self-hosting static assets

### LINKED ISSUE #305: Document self-hosting static assets
Document our decision to self-host static assets like jQuery, Mathjax, etc.

josebolos:
Mostly inspired by Harry Roberts' post: https://csswizardry.com/2019/05/self-host-your-static-assets/

Fixes #305


practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+
+By **static assets** we mean webfonts, JavaScript libraries (like jQuery or Mathjax), and any other files that get loaded by the browser from a domain that is not ours. These static assets are characterised by:
+
+* Being essential or important for the working of a page.
+* Being updated infrequently.
+* Require some manual intervention in order to be updated.
+
+Usually these assets are added to the page as part of the application, and not through tag managers.
+
+For this kind of static resources is our choice to **self-host them in our servers** instead of including them on our pages from an external CDN.
```
jpw:
```suggestion
For this kind of static resources is our choice to **self-host them on our servers** instead of including them on our pages from an external CDN.
```

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
+# Managing static assets
+
+By **static assets** we mean webfonts, JavaScript libraries (like jQuery or Mathjax), and any other files that get loaded by the browser from a domain that is not ours. These static assets are characterised by:
+
+* Being essential or important for the working of a page.
+* Being updated infrequently.
+* Require some manual intervention in order to be updated.
```
jpw: Not sure of this characterisation, as some will be made by build tools... what is the intention here?

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages.
+* **Consistency.** Static assets are cached and served in exactly the same way that the reset of our resources are, and they use the same CDN. As a result, timings and user experience is more consistent all around the world.
+* **Risk** that the service hosting the asset shuts down in the future, forcing us to upgrade a bunch of apps, some of which may be unmaintained.
+* **Performance.** Self-hosting can save us the cost of the DNS request, TCP connection and TLS negotiation, which in slow connections can easily be between 1000ms and 2000ms. This can also be mitigated with `rel=preconnect`.
+* **Filtering.** Some CDNs are filtered by the big firewall in China. This can slow down the loading of the page, or even make them break. Self-hosting static assets prevents this.
+* **Security.** Self hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) in the future.
```
jpw:
```suggestion
* **Security.** Self-hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) in the future.
```

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+
+Usually these assets are added to the page as part of the application, and not through tag managers.
+
+For this kind of static resources is our choice to **self-host them in our servers** instead of including them on our pages from an external CDN.
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages.
+* **Consistency.** Static assets are cached and served in exactly the same way that the reset of our resources are, and they use the same CDN. As a result, timings and user experience is more consistent all around the world.
+* **Risk** that the service hosting the asset shuts down in the future, forcing us to upgrade a bunch of apps, some of which may be unmaintained.
```
jpw:
```suggestion
* **Security.** reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface).
```

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+* Being updated infrequently.
+* Require some manual intervention in order to be updated.
+
+Usually these assets are added to the page as part of the application, and not through tag managers.
+
+For this kind of static resources is our choice to **self-host them in our servers** instead of including them on our pages from an external CDN.
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages.
```
jpw:
```suggestion
* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages, or closure of the service.
```

practices/managing-static-assets.md
jpw: This "Risk" point probably covered by slightly expanding the first point in this list, as per following suggestion.

practices/managing-static-assets.md
josebolos: The reason for creating the document is to address the problem that we've seen recently with some resources being blocked from China, so this document is specific to the third party resources that are let's say "first class citizens" of our sites and are required in order to get part of the functionality. Things like webfonts, jQuery, MathJax, and any other libraries we may have.

By "built tools" I think that you mean concatenated JS and CSS files? If that's the case, I wasn't thinking about these while writing this document, I'm assuming that we always serve those from our own servers.

jpw: I was thinking more of sprite sheets and other UI images. I guess it's the extent of the "manual intervention" that's confusing me...

josebolos: I see it's not very clear in my explanation. By "manual intervention" I'm thinking of instances where the app links to `${CDN}/jquery-3.2.0.min.js` (and not to `${CDN}/jquery-latest.min.js`), so in order to upgrade a dev has to replace the link destination with `${CDN}/jquery-3.2.1.min.js` or whatever.

So, not a result of a build process, not a result of an automated tool.

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+
+By **static assets** we mean webfonts, JavaScript libraries (like jQuery or Mathjax), and any other files that get loaded by the browser from a domain that is not ours. These static assets are characterised by:
+
+* Being essential or important for the working of a page.
+* Being updated infrequently.
+* Require some manual intervention in order to be updated.
+
+Usually these assets are added to the page as part of the application, and not through tag managers.
+
+For this kind of static resources is our choice to **self-host them on our servers** instead of including them on our pages from an external CDN.
```
danyalaytekin: is our choice -> it is our choice?

practices/managing-static-assets.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages, or closure of the service.
+* **Consistency.** Static assets are cached and served in exactly the same way that the reset of our resources are, and they use the same CDN. As a result, timings and user experience is more consistent all around the world.
```
danyalaytekin: is -> are?

practices/managing-static-assets.md
jpw: Aha ok, well the example you cite of webfonts are I think an example of where manual intervention is not required. So we could say "Possibly require some manual intervention..." or just drop the point, as I'm not sure how relevant it is to the point we are trying to make? Or am I missing something?

josebolos:
Just updated the introduction text which has now a simpler description of static assets.

josebolos:
@jpw Mind relooking at it? 💃

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages, or closure of the service.
+* **Consistency.** Static assets are cached and served in exactly the same way that the reset of our resources are, and they use the same CDN. As a result, timings and user experience are more consistent all around the world.
+* **Security.** reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface).
+* **Performance.** Self-hosting can save us the cost of the DNS request, TCP connection and TLS negotiation, which in slow connections can easily be between 1000ms and 2000ms. This can also be mitigated with `rel=preconnect`.
+* **Filtering.** Some CDNs are filtered by the big firewall in China. This can slow down the loading of the page, or even make them break. Self-hosting static assets prevents this.
+* **Security.** Self-hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) in the future.
```
jpw: Should this form a sub-list under the other Security point, above? Also suggest removing "in the future"

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
...
+
+Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts.
+
+This content rarely changes and its contents don't depend on user input or preferences.
+
+For this kind of static resources it is our choice to **self-host them on our servers** instead of including them on our pages from an external CDN.
+
+## Benefits of self-hosting static assets
+
+* **Resilience.** As long as our site is up, the resource will be up. We don't depend on other servers or additional CDNs and we won't be affected by their outages, or closure of the service.
```
dotcode, line 11: Resilience - as I understand it - is:

> an outcome of a highly automated, well architected and well tested system, predicated on the following common characteristics:    
> 
> - **Adaptive** — the ability to adapt to changing conditions.
> - **Self Healing** — the ability to recover from failures autonomously.
> - **Predictable** — the ability to consistently define responses to failures. This is a prerequisite to support the automation necessary to be adaptive and self healing.

If we are suggesting some method of including static assets on our servers as _a fallback_ for when the CDN fails, then that would be considered a resilient approach.

If however we're proposing moving stuff from a CDN to our own servers, is that not making it more fragile? Do we have better uptime than our CDN?

dotcode: Ah hold on … do you mean "only host on Fastly", rather than "don't host on _any_ CDN"?

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
...
+* **Consistency.** Static assets are cached and served in exactly the same way that the reset of our resources are, and they use the same CDN. As a result, timings and user experience are more consistent all around the world.
```
dotcode, line 12:
```suggestion
* **Consistency.** Static assets are cached and served like the rest of our resources are, and they use the same CDN. As a result, timings and user experience are more consistent all around the world.
```

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
...
+* **Security.** reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface).
```
dotcode, line 13: Can you give a one-liner as to how?

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
...
+* **Performance.** Self-hosting can save us the cost of the DNS request, TCP connection and TLS negotiation, which in slow connections can easily be between 1000ms and 2000ms. This can also be mitigated with `rel=preconnect`.
```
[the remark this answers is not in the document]
dotcode:
```suggestion
* **Performance.** Self-hosting can save us the cost of the DNS request, TCP connection and TLS negotiation, which in slow connections can easily be between 1000ms and 2000ms.
```

This gives the impression to someone reading that they could instead use rel-preconnect, and ignore the other advice. Is that what we want? Maybe add it back, but instead say something to qualify it along the lines of "This _can_ be mitigated with `rel=preconnect`, but … yadda yadda yadda"?

dotcode (changes requested):
Nice work Jose! Just a few questions/comments.

practices/managing-static-assets.md
```diff
@@ -0,0 +1,16 @@
+# Managing static assets
+
+Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts.
+
+This content rarely changes and its contents don't depend on user input or preferences.
```
danyalaytekin, line 5: 👋 

```suggestion
This content rarely changes and doesn't depend on user input or preferences.
```

?

practices/managing-static-assets.md
josebolos: > Ah hold on … do you mean "only host on Fastly", rather than "don't host on _any_ CDN"?

I mean "hosting jQuery ourselves" vs "loading jQuery from https://code.jquery.com/". What would be the appropriate word to describe that beyond "resilience"?

josebolos:
Rewrote the whole thing in order to include all of your suggestions.

practices/managing-static-assets.md
```diff
@@ -0,0 +1,13 @@
...
+
+We host on our own infrastructure these kinds of static resources, instead of including them on our pages from an third-party CDN. This applies especially to JavaScript files or libraries, and font files.
+
+Self-hosting static assets has several advantages over relying on third parties for their delivery:
+
+* **Resilience.** As long as our site is up, the static resource will be up too. We don't depend on third party servers or additional CDNs besides our own, and we won't be affected by their outages, or closure of the service.
+* **Consistency.** Static assets are cached and served like the rest of our resources are, and they use the same CDN. As a result, resource timings and user experience are more consistent all around the world.
+* **Security.** reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface) as the resources will enjoy the same level of security as the rest of first party content, and we don't depend on third parties being secure. Self-hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) for an additional layer of security.
```
danyalaytekin, line 11:
```suggestion
* **Security.** Reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface) as the resources will enjoy the same level of security as the rest of first party content, and we don't depend on third parties being secure. Self-hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) for an additional layer of security.
```

josebolos: Good catch! Thanks, @danyalaytekin!

practices/managing-static-assets.md
```diff
@@ -1,16 +1,13 @@
 # Managing static assets
 
-Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts.
+Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts. Static assets change rarely, and their content doesn't depend on user input or preferences.
 
-This content rarely changes and its contents don't depend on user input or preferences.
+We host on our own infrastructure these kinds of static resources, instead of including them on our pages from an third-party CDN. This applies especially to JavaScript files or libraries, and font files.
```
alexkilgour, line 5: "We host these kinds of static resources on our own infrastructure" instead of "We host on our own infrastructure these kinds of static resources"?

practices/managing-static-assets.md
```diff
@@ -1,16 +1,13 @@
...
+* **Consistency.** Static assets are cached and served like the rest of our resources are, and they use the same CDN. As a result, resource timings and user experience are more consistent all around the world.
+* **Security.** reduces our front-end [attack surface](https://en.wikipedia.org/wiki/Attack_surface) as the resources will enjoy the same level of security as the rest of first party content, and we don't depend on third parties being secure. Self-hosting the asset makes it easier for us to use [Subresource Integrity (SRI)](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) for an additional layer of security.
+* **Performance.** Self-hosting can save the users the cost of the DNS request, TCP connection and TLS negotiation for each self-hosted request, which in slow connections can make a big difference. This can also be mitigated with [`rel=dns-prefetch`](https://developer.mozilla.org/en-US/docs/Learn/Performance/dns-prefetch) and [`rel=preconnect`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preconnect), but self-hosting the assets eliminates that penalty completely.
```
alexkilgour, line 12: "Self-hosting can save the users the cost of the DNS request, TCP connection and TLS negotiation for each self-hosted request, which in slow connections can make a big difference."

Should this say "Self-hosting can save the users the cost of the DNS request, TCP connection and TLS negotiation for each **CDN-hosted** request, which in slow connections can make a big difference."

The cost of those requests is on the CDN, and we save those cost _by_ self-hosting?

practices/managing-static-assets.md
```diff
@@ -0,0 +1,13 @@
+# Managing static assets
+
+Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts. Static assets change rarely, and their content doesn't depend on user input or preferences.
+
+We host on our own infrastructure these kinds of static resources, instead of including them on our pages from an third-party CDN. This applies especially to JavaScript files or libraries, and font files.
```
jpw, line 5:
```suggestion
We host these kinds of static resources on our own infrastructure & CDN, instead of including them on our pages from other peoples infrastructure & CDN. This applies especially to JavaScript files or libraries, and font files.
```

I think this is an important clarification -- at first read I understood this line to mean "do not host on any CDNs (even those we control)"

practices/managing-static-assets.md
```diff
@@ -0,0 +1,13 @@
...
+* **Performance.** Self-hosting can save the users the cost of the DNS request, TCP connection and TLS negotiation for each self-hosted request, which in slow connections can make a big difference. This can also be mitigated with [`rel=dns-prefetch`](https://developer.mozilla.org/en-US/docs/Learn/Performance/dns-prefetch) and [`rel=preconnect`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preconnect), but self-hosting the assets eliminates that penalty completely.
+* **Filtering.** Some CDNs are filtered by the big firewall in China. This can slow down the loading of certain resources, or even make them fail to load completely. Self-hosting static assets helps mitigate this, as all of the 1st party resources on a page will be served from the same place.
```
jpw, line 13:
```suggestion
* **Filtering.** Some CDNs are filtered by China's "Great Firewall". This can slow down the loading of certain resources, or even make them fail to load completely. Self-hosting static assets helps mitigate this, as all of the 1st party resources on a page will be served from the same place.
```

jpw (changes requested):

practices/managing-static-assets.md
josebolos: Doesn't "third-parties infrastructure and CDN" sound better than "other peoples infrastructure & CDN"? Especially considering that joke about "the cloud is just other people's computers"?

I'm wondering how can I make it clear that I mean "our server and CDN" versus "random-js-library-cdn.com" without having to mention any of them by name.

jpw: i dont think so no, "our" and "others" are opposites.

practices/managing-static-assets.md
jpw: Does it not only save DNS, not TCP & TLS?

josebolos: I guess that will depend if it's a different subdomain and the cert covers all the (sub)domains, or a completely separate domain, but I don't think it makes sense to complicate the explanation so much.

josebolos:
_I think_ I got everyone's comments this time.

jpw (approved):

practices/managing-static-assets.md
```diff
@@ -0,0 +1,13 @@
+# Managing static assets
+
+Static assets are files that don't need to be generated, modified or processed by an application before being delivered to the user. This typically includes CSS, JavaScript, images and web fonts. Static assets change rarely, and their content doesn't depend on user input or preferences.
+
+We host these kinds of static resources on our own infrastructure and Content Delivery Network (CDN) instead of including them on our pages from a third-party CDN. This applies especially to JavaScript files or libraries, and font files.
+
+Self-hosting static assets has several advantages over relying on third parties for their delivery:
+
+* **Resilience.** As long as our site is up, the static resource will be up too. We don't depend on third party servers or additional CDNs besides our own, and we won't be affected by their outages, or closure of the service.
```
dotcode, line 9:
```suggestion
* **Control.** As long as our site is up, the static resource will be up too. We don't depend on third party servers or additional CDNs besides our own, and we won't be affected by their outages, or closure of the service.
```
This is not to do with _resiliency_. We are simply swapping one point of potential failure (someone' else's server) for another one (our own server). This is instead to do with _control_.

A _resilient_ solution would be (for instance), falling back to our own servers were the 3rd party ones down.

dotcode (changes requested):

dotcode (approved):

PR #316, opened 2019-07-29
Add skip links guide

sonniesedge:
We've had a team asking how to implement Skip Navigation links. So here's a short guide and link to resources.

### DIFF
--- a/accessibility/accessibility-checklist.md
+++ b/accessibility/accessibility-checklist.md
@@ -43,7 +43,7 @@ Resources:
 
 Not all users are able to navigate by landmarks. For those users the multiple links typically found at the head of a document all have to be read via assistive technology, or laboriously tabbed through using a keyboard.
 
-To help them out you can add a "Skip Navigation" link to the top of each page which, when clicked, will take them to the main section of the page.
+To help them out you can add a "Skip Navigation" link to the top of each page which, when activated, will take them to the main section of the page.
 
 We do this:
 
--- a/accessibility/accessibility-checklist.md
+++ b/accessibility/accessibility-checklist.md
@@ -41,7 +41,7 @@ Resources:
 
 ### Provide skip links to bypass repetitive content
 
-Not all users are able to navigate by landmarks. For those users the multiple links typically found at the head of a document all have to be read via assistive technology, or laboriously tabbed through using a keyboard.
+Not all users are able to easily navigate by landmarks. Multiple navigation links typically found at the top of a document all have to be laboriously tabbed through using a keyboard, on every page the user visits.
 
 To help them out you can add a "Skip Navigation" link to the top of each page which, when activated, will take them to the main section of the page.
 
--- a/accessibility/accessibility-checklist.md
+++ b/accessibility/accessibility-checklist.md
@@ -39,6 +39,30 @@ Resources:
 * [Tool: HTML5 accessibility feature support for sections](http://www.html5accessibility.com/#sections)
 
 
+### Provide a Skip Navigation option
+
+Not all users are able to navigate by landmarks. For those users the multiple links typically found at the head of a document all have to be read via assistive technology, or laboriously tabbed through using a keyboard.
+
+To help them out you can add a "Skip Navigation" link to the top of each page which, when clicked, will take them to the main section of the page.
+
+We do this:
+
+```
+<body>
+<p><a href="#maincontent">Skip to main content</a></p>
+...
+<main id="maincontent">
+<h1>Heading</h1>
+<p>This is the first paragraph</p>
+```
+
+You should supply CSS that hides the Skip Navigation _for sighted users only_, and makes it highly visible when it receives focus via tabbing.
+
+Resources:
+
+* [https://webaim.org/techniques/skipnav/](WebAIM's overview of Skip Navigation techniques)
+* [https://link.springer.com](Springer Link: an example of a well-done skip navigation)
+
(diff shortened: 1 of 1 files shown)


accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
 * [Tool: HTML5 accessibility feature support for sections](http://www.html5accessibility.com/#sections)
 
 
+### Provide a Skip Navigation option
```
hollsk:
```suggestion
### Provide skip links to bypass repetitive content
```

Suggesting this instead of "skip navigation", because there are other things you might want to add a bypass to (e.g. very long, tiresome widgets with 100s of tab stops)

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+
+Not all users are able to navigate by landmarks. For those users the multiple links typically found at the head of a document all have to be read via assistive technology, or laboriously tabbed through using a keyboard.
```
hollsk:
```suggestion
Not all users are able to easily navigate by landmarks. Multiple navigation links typically found at the top of a document all have to be laboriously tabbed through using a keyboard, on every page the user visits.
```

Removed the bit about assistive tech just coz everything needs to be read by AT anyway. That changes the focus to how much tedious work it is to physically tab through everything. Also added a bit about how it needs to be done over and over on every page to make it even more 😩

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+
+To help them out you can add a "Skip Navigation" link to the top of each page which, when clicked, will take them to the main section of the page.
```
hollsk:
```suggestion
To help them out you can add a "Skip Navigation" link to the top of each page which, when activated, will take them to the main section of the page.
```

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+
+We do this:
+
+```
+<body>
+<p><a href="#maincontent">Skip to main content</a></p>
+...
+<main id="maincontent">
```
hollsk: One thing that Link do is to add a `tabindex="-1"` to the `main`, which allows it to more consistently receive focus when using the skip link. I feel like we probably need to have that in the example (as not having it will cause bugs), but I'm wary of having both too many explanations, or of having much stuff in the example that isn't explained 🤔 

I wonder if skip links would be a good candidate for a component? Then we could link off to the component at a later date instead of explaining all the bits

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+<h1>Heading</h1>
+<p>This is the first paragraph</p>
+```
+
+You should supply CSS that hides the Skip Navigation _for sighted users only_, and makes it highly visible when it receives focus via tabbing.
```
hollsk: Might be worth adding a bit here about how you can also use skip links to bypass other annoying things here (there's a bit in the WebAIM article about that, under the "multiple skip links" section)

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+
+Resources:
+
+* [https://webaim.org/techniques/skipnav/](WebAIM's overview of Skip Navigation techniques)
```
hollsk:
```suggestion
* [Article: WebAIM's overview of Skip Navigation techniques](https://webaim.org/techniques/skipnav/)
```

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,30 @@ Resources:
...
+* [https://link.springer.com](Springer Link: an example of a well-done skip navigation)
```
hollsk:
```suggestion
* [Example: Springer Link's well-done skip navigation](https://link.springer.com)
```

accessibility/accessibility-checklist.md
sonniesedge: pls endorse me on linkedin for markdown

accessibility/accessibility-checklist.md
sonniesedge: Good idea. Added a wee bit about that.

accessibility/accessibility-checklist.md
sonniesedge: I think that potentially confuses things in what is a very brief section. That WebAIM section itself becomes incredibly nuanced, as they suggest _not_ adding more skip links, and instead using better document structure and in page navigation.

accessibility/accessibility-checklist.md
sonniesedge: And yeah, I think this is an excellent candidate for a component!

accessibility/accessibility-checklist.md
```diff
@@ -39,6 +39,33 @@ Resources:
...
+```
+<body>
+<p><a href="#maincontent">Skip to main content</a></p>
+...
+<main id="maincontent" tabindex="-1">
+<h1>Heading</h1>
+<p>This is the first paragraph</p>
+```
+
+We add `tabindex="-1"` to the targeted element so that it can be visibly focussed when skipped to.
```
hollsk:
```suggestion
We add `tabindex="-1"` to the targeted element so that it can be programmatically focussed when skipped to.
```
😉

PR #320, opened 2019-08-12
Add section about forms and submit buttons

### LINKED ISSUE #200: Document approach to form submit buttons
There was recently a discussion on Slack wrt whether to use `<input>` or <button>` for form submit buttons. It was agreed in the end that we should use `<button>`.

See #frontend 26th Sept 1:54PM: https://springernature.slack.com/archives/C0GDUEGMS/p1506430489000133

Someone needs to formalise this in the playbook as an agreed approach including rationale.

josebolos:
Fixes #200.

### DIFF
--- a/markup/house-style.md
+++ b/markup/house-style.md
@@ -9,6 +9,7 @@ This document outlines the way we write markup and why. See our [house style doc
 - [Semantics](#semantics)
 	- [Follow the HTML 4 outline model for heading levels](#follow-the-html-4-outline-model-for-heading-levels)
 - [Templates](#templates)
+- [Forms](#forms)
 
 ## HTML vs XHTML
 
@@ -121,3 +122,23 @@ We *don't* do this:
 ## Templates
 
 Templates should, where possible, be written using [DustJS](http://www.dustjs.com/) or [Handlebars](http://handlebarsjs.com/). Use of linting is encouraged to prevent common errors. For projects using DustJS  [Dustmite](https://www.npmjs.com/package/dustmite) should be used. Dustmite will prevent any syntax errors. It also enforces consistent style in your template code and can catch more subtle errors like usage of the potentially unsafe `{@if}` dust helper.
+
+## Forms
+
+On modern browsers, `button type="submit"` provides a superset of functionality when compared to `input type=submit`.
+
+We do this:
+
+```html
+<form>
+	<button type="submit">Search</button>
+</form>
+```
+
+We *don't* do this:
+
+```html
+<form>
+	<input type="submit">Search</input>
+</form>
+```


markup/house-style.md
```diff
@@ -121,3 +122,23 @@ We *don't* do this:
 ## Templates
 
 Templates should, where possible, be written using [DustJS](http://www.dustjs.com/) or [Handlebars](http://handlebarsjs.com/). Use of linting is encouraged to prevent common errors. For projects using DustJS  [Dustmite](https://www.npmjs.com/package/dustmite) should be used. Dustmite will prevent any syntax errors. It also enforces consistent style in your template code and can catch more subtle errors like usage of the potentially unsafe `{@if}` dust helper.
+
+## Forms
+
+On modern browsers, `button type="submit"` provides a superset of functionality when compared to `input type=submit`.
```
morgaan: Interesting! Could you join a resource that list the superset of functionality. I only know that you can then nest other elements inside it. In "Form Design Pattern" it is stated that this variant have the side effect in older version of IE to submit the value of all the buttons to the server, regardless of which one was clicked, so you'll need to know which button was [clicked](http://smashed.by/submitbuttons) so you can determine the right course of action to take. Maybe this is worth mentioning.

josebolos: Hi @morgaan - I just offered to write this so we could rid of an issue that has been open for two years. To be honest, I'm not familiar with the implications, I was just summarising a conversation that we had in Slack years ago.

If you're more familiar with this than me, feel free to amend the text that I've submitted!

morgaan: Hi @josebolos - First of all apologies that I did not mentioned it was an existing issue pending for long time. I thought it was coming out from work you being busy with lately.

I do not know much more TBH, I just remembered these lines from the book and I have been using mostly the input variant over the years, unless I had the need to nest stuffs in there. This is what is maybe meant by superset of functionality.

IMO I would let it open to developers to pick from, depending on their use case.
Reading the "We don't do this" looks like the input variant is bad practice which I doubt it is, at least according to Adam Silver, author of the book.
This may though be worth mentioning that in case of using the `button` variant that it does not fall into the edge cases category.

What do you think? We could also bring this topic into the next FE open space to see what is the fellowship opinion on that matter 2 years later.

In any case thank you for offering to write this ☀️. I plan to come up with a draft of the few tips I got from the book at writing forms, and already applied to NatCar project. This one could end up being a guideline chapter into the Playbook about writing forms and or a part in the Elements design system styleguide.

jpw: @morgaan sounds great, thanks very much!
I also vaguely remember @davidpauljunior having some thoughts on button vs input, do you have some David? :)

markup/house-style.md
```diff
@@ -121,3 +122,23 @@ We *don't* do this:
...
+<form>
+	<button type="submit">Search</button>
+</form>
+```
+
+We *don't* do this:
+
+```html
+<form>
+	<input type="submit">Search</input>
```
axemonkey: We _definitely_ don't do that, since `<input>` is a [self-closing tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Technical_summary). 😉

morgaan: Indeed, well spotted 🏅

josebolos:
angry women yelling at cat

Done: https://github.com/springernature/frontend-open-space/issues/240

markup/house-style.md
hollsk: >older version of IE

Do we know which older version of IE? If it's old enough that it can't support TLS 1.2 then it won't be clicking any buttons on any of our sites because it won't be able to access them at all: https://github.com/springernature/frontend-playbook/blob/master/practices/graded-browser-support.md#browsers-without-tls-12-support

morgaan: Still in IE11 as far as I understand. But it also seems that you should be safe if you provide the `button` with  `name` and `value` attributes. In such case we should emphase this in the Playbook if using the `button` variant.

PR #321, opened 2019-08-19
Forms checklist

sonniesedge:
Some checklists to follow when building forms. 

A work in progress, so don't expect it to be anywhere near perfect just yet. But happy to receive comments even at this stage.

### DIFF
--- a/markup/forms.md
+++ b/markup/forms.md
@@ -1,0 +1,71 @@
+# Forms
+
+## Checklist
+
+### Buttons rather than inputs
+
+- Specify the required attributes (name and value)
+
+### Multistep forms over multiple submit buttons
+
+### Server-side Validation
+
+- Start with server-side validation
+- Return an error message for each erroring form element.
+
+### Client-side validation
+
+### Accessibility
+
+### Label every input
+
+### Use the most simple form control that does the job
+
+### Minimise the number of fields to fill out
+
+### Mark optional fields
+
+### Size according to the expected input
+
+### Retain focus for every input element
+
+- For preference, enhance it! Make it part of the brand!
+
+### Don't ask users to repeat their email address
+
+### Provide a "show password" option
+
+### Don't split data fields
+
+### Avoid dropdown menus
+
+### Provide examples of input
+
+- Don't use placeholders!
+
+### Offer help text
+
+### Masked input
+
+### Use appropriate input types
+
+### Explain clearly why you need any private information
+
+### Use autocomplete when useful
+
+### Enable auto-filling of personal details
+
+### Use fieldsets and legends
+
+### Make use of native validation, enhance when necessary
+
+### Avoid native datepickers
+
+- Controversial.
+- Often better to stick with date fields.
+
+## Resources
+
+- https://www.smashingmagazine.com/2018/08/best-practices-for-mobile-form-design/
+- https://github.com/cferdinandi/bouncer
+- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints


markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Accessibility
+
+### Label every input
+
+### Use the most simple form control that does the job
+
+### Minimise the number of fields to fill out
+
+### Mark optional fields
```
morgaan: - Yes ☀️ it is preferred over the * (required) pattern
- If they are optional, it might be worse to re-evaluate if they are really needed after all and if they might not be getting in the way to conversion

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+### Server-side Validation
+
+- Start with server-side validation
+- Return an error message for each erroring form element.
+
+### Client-side validation
+
+### Accessibility
+
+### Label every input
```
morgaan: Do not hesitate to provide the most information possible so that the user succeed rather than get into validation issue. Not a good feeling to have.
Therefore you could divide the label into 2 parts, the label itself (e.g Date of birth) and a hint (e.g DD/MM/YYYY basically what you often find in placeholder).
A good practice is also to nest the inline error message inside of the label so that it will be spoken by Assistive tech when focused.

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Mark optional fields
+
+### Size according to the expected input
+
+### Retain focus for every input element
+
+- For preference, enhance it! Make it part of the brand!
+
+### Don't ask users to repeat their email address
```
morgaan: nor password (see "show password" below)

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Provide a "show password" option
```
morgaan: as an enhancement (with JS)

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Don't split data fields
+
+### Avoid dropdown menus
```
morgaan: Use radio buttons for single option and checkbox for multiple options when a decent amount of options are available (below 8)

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Provide examples of input
+
+- Don't use placeholders!
+
+### Offer help text
```
morgaan: Mostly as hint in labels as stated above

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Provide a "show password" option
+
+### Don't split data fields
+
+### Avoid dropdown menus
+
+### Provide examples of input
+
+- Don't use placeholders!
```
morgaan: Use label hints instead as stated above

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Offer help text
+
+### Masked input
```
morgaan: I thought this to be a good practice until I read something that challenged it. The reasons do not come out from the top of my head atm, therefore I would research that one. Though Estelle Weyl has a great [framework agnostic accessible input masking library](https://github.com/estelle/input-masking/)

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Use appropriate input types
+
+### Explain clearly why you need any private information
+
+### Use autocomplete when useful
```
morgaan: as an enhancement

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Enable auto-filling of personal details
+
+### Use fieldsets and legends
+
+### Make use of native validation, enhance when necessary
```
morgaan: Excerpt of the Form Design Pattern that may challenge it:

"While HTML5 validation support is quite good, it’s not implemented uniformly. For example, the required attri- bute can mark fields as invalid from the outset, which isn’t desirable. Some browsers, such as Firefox 45.7, will show an error of “Please enter an email address” even if the user entered something in the box, whereas Chrome, for example, says “Please include an ‘@’ in the email address,” which is more helpful.
We also want to give users the same interface whether errors are caught on the server or the client. For these rea- sons we’ll design our own solution. The first thing to do is turn off HTML5 validation:" using the novalidate attribute on the form element.

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+### Avoid native datepickers
+
+- Controversial.
+- Often better to stick with date fields.
+
+## Resources
+
+- https://www.smashingmagazine.com/2018/08/best-practices-for-mobile-form-design/
+- https://github.com/cferdinandi/bouncer
+- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints
```
morgaan: - https://adamsilver.io/articles/form-design-from-zero-to-hero-all-in-one-blog-post/

morgaan:
Hi Charlie, this looks already like quite a good draft. Love of points I would put ❤️ on. I was supposed to bring that contribution to the playbook, but being busy on NatCar, so I'm glad you fire started it. I added a few comments from things I learned from the Form Design Patterns book which I found relevant. Hope it's alright 😊

markup/forms.md
jpw: "... as a progressive enhancement" ?

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Retain focus for every input element
+
+- For preference, enhance it! Make it part of the brand!
+
+### Don't ask users to repeat their email address
+
+### Provide a "show password" option
+
+### Don't split data fields
```
jpw: What does this mean please?

markup/forms.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Masked input
+
+### Use appropriate input types
+
+### Explain clearly why you need any private information
+
+### Use autocomplete when useful
+
+### Enable auto-filling of personal details
```
jpw: Yes, and this might be useful to link to? 😁 https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md#a-note-on-autofill

jpw:
Great idea, lovely stuff. Might want to add a bit about sane default values? 😁 https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md#use-sane-form-defaults

markup/forms.md
morgaan: Good point! Thinking out loud too: Then maybe two password inputs in raw HTML. When JS available, then enhance to "show password" feature the first one, and turn the other one into a hidden input and sync what is typed in first one so that the BE/FE contract stays the same with or without JS. Sounds like a bit of work already, therefore it (or any better implementation) could qualifiy as a component candidate for the Design system/Global toolkit I believe

markup/forms.md
morgaan: For instance Date: [  ] / [  ] / [    ] (3 fields). Just a guess though

markup/forms.md
```diff
@@ -1,42 +1,128 @@
...
+
+### Avoid `<select>` when possible
+
+- `<select>` menus have accessibility and usability considerations.
+- Replace `<select>` with `<input type='radio'>` whenever possible.
+- If you have too many options for radio buttons, then consider how you could better present the data.
 
 ### Mark optional fields
 
+- Highlight form fields that are _optional_, rather than highlighting fields that are required.
```
jpw: Yes, but why? Can we link out to any research?

markup/forms.md
```diff
@@ -1,42 +1,128 @@
...
 
-### Client-side validation
+- Part of the Core experience, so every form should be initially built to the Core standards.
+  - Data should be validated on the server, and a form with inline messages for any errors encountered.
+- All data being sent to a server must be sanitised, so server-side validation is always necessary.
 
-### Accessibility
+### Allow client-side validation to occur
+
+- Client-side validation is a supplement to server-side validation, not a replacement.
```
jpw: "Client-side validation is a progressive enhancement of server-side validation" ?

markup/forms.md
```diff
@@ -1,42 +1,128 @@
...
+- Replace `<select>` with `<input type='radio'>` whenever possible.
+- If you have too many options for radio buttons, then consider how you could better present the data.
 
 ### Mark optional fields
 
+- Highlight form fields that are _optional_, rather than highlighting fields that are required.
+- Consider removing optional form fields: if they are optional then why is the data being collected?
+- Fewer fields means less cognitive work for the user!
+
+### Minimise the number of fields to fill out
```
jpw: this heading appears to relate to the previous line

markup/forms.md
hollsk: Also in the a11y checklist https://github.com/springernature/frontend-playbook/blob/master/accessibility/accessibility-checklist.md#help-users-make-the-most-of-autocomplete-and-autofill

PR #322, opened 2019-08-28
Add section about Form submit controls

morgaan:
Outcome of a frontend open space discussion

### DIFF
--- a/markup/house-style.md
+++ b/markup/house-style.md
@@ -85,8 +85,12 @@ inner HTML content (think `<em>`, `<strong>`, or even `<img>`), and use
 `::after` and `::before` pseudo-elements to achieve complex rendering while
 `input` only accepts a textual value attribute.
 
+**Note**: Submitting is `button` default behaviour, but we want to "code for
+humans". Therefore do not ommit `type="submit"`.
+
 Make sure to specify `name` and `value` attributes to maximize browsers
-compatibility.
+compatibility ([See IE11-Only Submit Bug on Stack
+Overflow](https://stackoverflow.com/a/22703881/9461391)).
 
 We do this:
 
@@ -96,13 +100,21 @@ We do this:
 </form>
 ```
 
-or
-
 We *don't* do this:
 
 ```html
 <form>
 	<input type="submit" value="Continue">
+</form>
+```
+
+and
+
+We *don't* do this:
+
+```html
+<form>
+	<button>Continue</button>
 </form>
 ```
 
@@ -114,17 +126,18 @@ This becomes problematic when using multiple submit buttons in a single form wit
 Here are some suggestions to work around these:
 
 - Interfaces could be redesigned so multiple submit buttons are not required.
-  - Replacing them with radio buttons and a single submit button (requires an
-    extra click from the user though);
-  - Use a mutliple pages form as recommended by Gov.uk in [Structuring forms,
+  - (Preferred) Use a multiple pages form as recommended by GOV.UK in [Structuring forms,
     Start with one thing per
     page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page).
+  - Replace them with radio buttons and a single submit button (requires an
+    extra click from the user though);
 - A separate form can be used for each instance, with a hidden `input` providing
   the data the submit `button` would normally carry. This can be a good
   solution when you have a simple “Delete this row” problem.
 - Last resort solution: Hide the value inside the name of the control. A loop
   over the names from the form data is needed in the business logic then. Please
-  avoid as much as you can!
+  avoid as much as you can as this adds complexity to code and may degrade
+  performance.
 
 ## Semantics
 
--- a/markup/house-style.md
+++ b/markup/house-style.md
@@ -6,6 +6,7 @@ This document outlines the way we write markup and why. See our [house style doc
 	- [Validation](#validation)
 	- [Do not omit optional end tags](#do-not-omit-optional-end-tags)
 	- [Implicitly close void elements](#implicitly-close-void-elements)
+	- [Form submit controls](#form-submit-controls)
 - [Semantics](#semantics)
 	- [Follow the HTML 4 outline model for heading levels](#follow-the-html-4-outline-model-for-heading-levels)
 - [Templates](#templates)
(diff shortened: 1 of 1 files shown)


markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+element. Indeed it submits the text within the `button` element as its value in
+the form data rather than the value of its `value` attribute.
+This becomes problematic when using multiple submit buttons in a single form with each a different value and purpose. 
+
+Here are some suggestions to work around these:
+
+- Interfaces could be redesigned so multiple submit buttons are not required.
+  - Replacing them with radio buttons and a single submit button (requires an
+    extra click from the user though);
+  - Use a mutliple pages form as recommended by Gov.uk in [Structuring forms,
```
dotcode:
```suggestion
  - Use a multiple pages form as recommended by GOV.UK in [Structuring forms,
```

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+### Form submit controls
+
+On modern browsers, `button type="submit"` provides a superset of functionality when compared to `input type=submit`.
+
+`button` elements are much easier to style than `input` elements. You can add
+inner HTML content (think `<em>`, `<strong>`, or even `<img>`), and use
+`::after` and `::before` pseudo-elements to achieve complex rendering while
+`input` only accepts a textual value attribute.
+
+Make sure to specify `name` and `value` attributes to maximize browsers
```
dotcode: How does this maximize browser compatibility? Is it possible to explain, or to link out to something?

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+or
+
+We *don't* do this:
+
+```html
+<form>
+	<input type="submit" value="Continue">
+</form>
+```
+
```
dotcode: Should we also add this (submitting is the default behaviour, but we want to "code for humans")?

We do this:

```html
<form>
	<button type="submit" …
</form>
```

We *don't* do this:

```html
<form>
	<button …
</form>
```

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+We do this:
+
+```html
+<form>
+	<button type="submit" name="myButton" value="foo">Continue</button>
+</form>
+```
+
+or
```
dotcode:
```suggestion
```

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+
+**Caution**: There is a shortcoming for Internet Explorer when using the `button`
+element. Indeed it submits the text within the `button` element as its value in
+the form data rather than the value of its `value` attribute.
+This becomes problematic when using multiple submit buttons in a single form with each a different value and purpose. 
+
+Here are some suggestions to work around these:
+
+- Interfaces could be redesigned so multiple submit buttons are not required.
+  - Replacing them with radio buttons and a single submit button (requires an
```
dotcode:
```suggestion
  - Replace them with radio buttons and a single submit button (requires an
```

markup/house-style.md
dotcode: Is this method is our preference (in most cases)? If so, let's highlight that fact - maybe putting it as the first option and stating it's our preference?

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+    extra click from the user though);
+  - Use a mutliple pages form as recommended by Gov.uk in [Structuring forms,
+    Start with one thing per
+    page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page).
+- A separate form can be used for each instance, with a hidden `input` providing
+  the data the submit `button` would normally carry. This can be a good
+  solution when you have a simple “Delete this row” problem.
+- Last resort solution: Hide the value inside the name of the control. A loop
+  over the names from the form data is needed in the business logic then. Please
+  avoid as much as you can!
```
dotcode: Let's state why we want to avoid this.

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
 ```html
 <meta charset="UTF-8" />
 ```
+### Form submit controls
+
+On modern browsers, `button type="submit"` provides a superset of functionality when compared to `input type=submit`.
```
hollsk:
```suggestion
In modern browsers, `button type="submit"` provides a superset of functionality when compared to `input type=submit`.
```

😉

markup/house-style.md
hollsk: There are funky line breaks in the markdown so I can't make the suggestion in the Github interface, but I would compress the first two sentences to this:

>**Caution**: In Internet Explorer,  the text within the `button` element is submitted as the value in the form data, rather than the value of the `value` attribute.

Also, I've already forgotten the details 😆 but is this just _some_ versions of IE? Might be worth mentioning that it's not all versions if that's the case

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+```
+
+**Caution**: There is a shortcoming for Internet Explorer when using the `button`
+element. Indeed it submits the text within the `button` element as its value in
+the form data rather than the value of its `value` attribute.
+This becomes problematic when using multiple submit buttons in a single form with each a different value and purpose. 
+
+Here are some suggestions to work around these:
+
+- Interfaces could be redesigned so multiple submit buttons are not required.
```
hollsk:
```suggestion
- Redesign the interface so multiple submit buttons are not required.
```
This suggestion changes the sentence from the passive voice to the active voice 😄

markup/house-style.md
```diff
@@ -75,6 +76,55 @@ We *don't* do this:
...
+  - Replacing them with radio buttons and a single submit button (requires an
+    extra click from the user though);
+  - Use a mutliple pages form as recommended by Gov.uk in [Structuring forms,
+    Start with one thing per
+    page](https://www.gov.uk/service-manual/design/form-structure#start-with-one-thing-per-page).
+- A separate form can be used for each instance, with a hidden `input` providing
```
hollsk:
```suggestion
- Use a separate form for each instance, with a hidden `input` providing
```
passive voice changed to active voice

morgaan:
Thank you so much @dotcode and @hollsk for the review. I believe I resolved everything. Therefor I ask for re-review.

morgaan:
@hollsk Would you mind re-review that one? It's pending for months :(

hollsk:
I think you need Jude to approve, though 🤔

morgaan:
> I think you need Jude to approve, though 🤔

Hi @dotcode any change you can give a quick look and hopefully approval to this one? Thank you ☀️

PR #325, opened 2019-09-06
Change misleading wording

dotcode:
Previous wording gave the impression we have achieved full compliance. Although in many areas we achieve very substantial compliance (far above the norms for the industry), we must not give the impression of full compliance where that does not exist.

### DIFF
--- a/accessibility/house-style.md
+++ b/accessibility/house-style.md
@@ -59,8 +59,7 @@ You may use [our VPAT template](https://github.com/springernature/vpat) when eva
 
 ## How we conform
 
-### Standards we conform to
-
+### Standards we aim for
 Our target standards are [WCAG 2.1](https://www.w3.org/TR/WCAG21/) level AA.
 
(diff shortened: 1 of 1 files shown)


dotcode:
Using privileges to squash and merge without review. Deal with it.

accessibility/house-style.md
```diff
@@ -59,8 +59,7 @@ You may use [our VPAT template](https://github.com/springernature/vpat) when eva
 
 ## How we conform
 
-### Standards we conform to
-
+### Standards we aim for
 Our target standards are [WCAG 2.1](https://www.w3.org/TR/WCAG21/) level AA.
```
josebolos: Needs an empty line between the title and the paragraph so markdownlint doesn't complain about it 😬

dotcode: Typical 🤦‍♂️

dotcode: On the upside, happily proving my own maxim that "there is no change too small for a review".

PR #331, opened 2019-10-21
A11y updates

hollsk:
I started this PR with the idea that I would rename the "house-style" document to "introduction", and review the info that we have already to structure it a bit more sensibly. But then I realised that we had a bunch of outdated references to various standards and legislations, and realised that it'd make sense to update those before i went any further. So! This PR:

* updates changes house-style.md to introduction.md 
* updates all the links to point to the new file

And then it:

* Changes the wording around the legislations we're obliged to meet so that the EU regs take precedence over everything else (they incorporate WCAG 2.1, so are more up-to-date)
* Makes the relationship between "EU Directive 2016/2102" and "EN 301 549" clear (they are the same thing)
* Makes it more clear that we aim to meet two separate levels of legislation; federal (like Section 508) and civil (like ADA) - previously we had those kind of mashed together

Then finally it:

* Updates the description of what a VPAT actually is so that it reflects the current update that incorporates EN 301 549
* Adds a link to the ITI website
* Makes the whole paragraph less US-oriented

I will proceed with the other changes I have in mind in a separate PR ✨


hollsk:
I realise that these commits are about as far from atomic as one could possibly get, and I am apologetic

PR #332, opened 2019-10-22
Add a note on autocomplete

### LINKED ISSUE #314: Add note on autofill to a11y checklist
On the [secure markup page](https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md) there's a note on autofill that goes over the reasons that we don't disable autofill for security reasons, and why we shouldn't autofill fields that are visually hidden. 

There are also accessibility reasons for not disabling the `autocomplete` functionality for any fields containing personal information - it can make it more difficult to use password managers or form fillers, which is a major negative for anyone whose memory isn't too good, or who struggles to type quickly into input fields (e.g. switch device users). 

We should add a note about that to the a11y checklist, and link back to the note on the secure markup page too.

hollsk:
resolves #314 

* Creates a new "forms" section that contains checklist items about forms
* Adds guidance about autocomplete/autofill from an accessibility perspective

### DIFF
--- a/accessibility/accessibility-checklist.md
+++ b/accessibility/accessibility-checklist.md
@@ -246,6 +246,26 @@ Resources:
 * [Article: 11 reasons why placeholders are problematic](https://medium.com/simple-human/10-reasons-why-placeholders-are-problematic-f8079412b960)
 * [Article: Don’t Use The Placeholder Attribute](https://www.smashingmagazine.com/2018/06/placeholder-attribute)
 
+### Help users make the most of autocomplete and autofill
+
+By default, browsers remember information that users submit into input fields. The browser can then suggest possible completions for fields that the user has started typing in (autocomplete) or pre-populate certain fields when the page loads (autofill). This behaviour can be turned off in HTML with the `autocomplete="false"` attribute.
+
+Don't indiscriminately disable autocomplete for input fields. The default autocomplete and autofill behaviours are useful for many users with and without disabilities. 
+
+For example, autocomplete helps users with motor disabilities by reducing the amount of typing they need to do in input fields. Autofill can help users with cognitive disabilities by automatically filling in common information like addresses or telephone numbers. 
+
+Don't disable autocomplete on sensitive fields like username and password fields. Any security benefits of doing this is negligible compared to the benefits of users' in-browser password managers, and has a direct negative impact on accessibility. Read the note on autofill in the [secure markup guide](https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md#a-note-on-autofill) for more on that. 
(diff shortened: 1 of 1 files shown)


jon-stevens:
image

accessibility/accessibility-checklist.md
```diff
@@ -228,6 +226,14 @@ Resources:
 * [Article: Opening links in a new window](https://www.nomensa.com/blog/2011/opening-links-new-window)
 
 
+## Forms
+
+### Labels and context
+
+Associate all form controls with appropriate labels.
```
alexkilgour: could be a 2 item list?

alexkilgour:
Looks good and makes sense. Do we want to expand on any other aspects of forms? for example, at the moment we are looking at issues such as (1) Should we use multi-stage forms (wizard style, over multiple pages), or (2) Should we use interactive forms (form fields update via JS based on previous answers)

accessibility/accessibility-checklist.md
```diff
@@ -240,6 +246,25 @@ Resources:
...
+
+Don't indiscriminately disable autocomplete for input fields. The default autocomplete and autofill behaviours are useful for many users with and without disabilities. 
+
+For example, autocomplete helps users with motor disabilities by reducing the amount of typing they need to do in input fields. Autofill can help users with cognitive disabilities by automatically filling in common information like addresses or telephone numbers. 
+
+Don't disable autocomplete on sensitive fields like username and password fields. Any security benefits of doing this is negligible compared to the benefits of users' in-browser password managers, and has a direct negative impact on accessibility. Read the note on autofill in the [secure markup guide](https://github.com/springernature/frontend-playbook/blob/master/security/secure-markup.md#a-note-on-autofill) for more on that.
```
jpw: <3

hollsk:
@alexkilgour 
>Looks good and makes sense. Do we want to expand on any other aspects of forms? for example, at the moment we are looking at issues such as (1) Should we use multi-stage forms (wizard style, over multiple pages), or (2) Should we use interactive forms (form fields update via JS based on previous answers)

Ooh, maybe. Do we have a firm conclusion? Without being part of those discussions, it looks to me like that has some nuance involved which might justify a fuller treatment than just one or two lines in the checklist?

alexkilgour:
> Ooh, maybe. Do we have a firm conclusion? Without being part of those discussions, it looks to me like that has some nuance involved which might justify a fuller treatment than just one or two lines in the checklist?

No firm conclusions, and yes in hindsight I think it is out of scope for this and needs more thought. But it is a subject that I think would benefit from some documented approaches as I think this would definitely help teams.

PR #334, opened 2019-11-13
Add a tools page

hollsk:
The accessibility introduction is more of an infodump than an introduction. This PR:

* Creates a new tools.md document
* Moves all the stuff about tools out of the introduction and into the tools doc
* Splits the tool types into sections depending on what they're trying to achieve
* Changes the suggestion of using Dictation on Mac to a suggestion of using Voice Control (released in MacOS Catalina)]
* Adds a really gigantic caveat to the disability simulator section 
* Adds an extra link to the disability simulator section

(Some of the additional content is copied from developing-accessible-products.md)

### DIFF
--- a/accessibility/introduction.md
+++ b/accessibility/introduction.md
@@ -103,3 +103,6 @@ The Government Digital Service frequently publishes excellent resources. Some hi
 * [We're making accessibility clearer and easier](https://gds.blog.gov.uk/2017/10/23/were-making-accessibility-clearer-and-easier/)
 * [Making your service accessible: an introduction](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction)
 
+#### Tools 
+
+A breakdown of the tools we recommend for testing accessibility, including assistive technology, can be found on the Playbook [Accessibility tools](tools.md) page. 
--- a/accessibility/introduction.md
+++ b/accessibility/introduction.md
@@ -10,6 +10,7 @@
     - [Guidelines](#guidelines)
     - [Testing with real users](#testing-with-real-users)
     - [Resources](#resources)
+    - [Tools](#tools)
 
 
 ## General principles
--- a/accessibility/introduction.md
+++ b/accessibility/introduction.md
@@ -6,7 +6,7 @@
 - [VPATs](#vpats)
 - [How we conform](#how-we-conform)
   - [Standards](#standards-we-aim-for)
-  - [Tools and techniques](#techniques)
+  - [Techniques](#techniques)
     - [Guidelines](#guidelines)
     - [Testing with real users](#testing-with-real-users)
     - [Resources](#resources)
--- a/accessibility/introduction.md
+++ b/accessibility/introduction.md
@@ -6,11 +6,8 @@
 - [VPATs](#vpats)
 - [How we conform](#how-we-conform)
   - [Standards](#standards-we-aim-for)
-  - [Tools and techniques](#tools-and-techniques)
+  - [Tools and techniques](#techniques)
     - [Guidelines](#guidelines)
-    - [Pa11y](#pa11y)
-    - [Other accessibility testing tools](#other-accessibility-testing-tools)
-    - [Assistive technology](#assistive-technology)
     - [Testing with real users](#testing-with-real-users)
     - [Resources](#resources)
 
@@ -70,7 +67,7 @@ As of January 18th 2018, WCAG 2.0 AA (the previous candidate recommendation) was
 Our expectation is that Section 508 will be updated to reference WCAG 2.1 in due course. Comply with WCAG 2.1 now to avoid retrofitting repairs later!  
 
 
-### Tools and techniques
+### Techniques
 
 #### Guidelines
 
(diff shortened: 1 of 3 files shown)


accessibility/introduction.md
```diff
@@ -6,11 +6,8 @@
 - [VPATs](#vpats)
 - [How we conform](#how-we-conform)
   - [Standards](#standards-we-aim-for)
-  - [Tools and techniques](#tools-and-techniques)
+  - [Techniques](#techniques)
     - [Guidelines](#guidelines)
-    - [Pa11y](#pa11y)
-    - [Other accessibility testing tools](#other-accessibility-testing-tools)
-    - [Assistive technology](#assistive-technology)
     - [Testing with real users](#testing-with-real-users)
```
alexkilgour: under `Testing with real users` there is now a `Tools` section that links off to the new tools page. Should this section be linked to from here?

accessibility/tools.md
```diff
@@ -0,0 +1,93 @@
...
+
+The GDS Accessibility team ran an experiment in 2017 to [audit the performance of a large number of available automated accessibility testing tools](https://accessibility.blog.gov.uk/2017/02/24/what-we-found-when-we-tested-tools-on-the-worlds-least-accessible-webpage/). They found that the highest number of errors that one tool was able to detect was 40%. The majority of the tools fell into the 20%-30% range. 
+
+
+### Pa11y and axe
+
+We use [Pa11y](http://pa11y.org/) to perform automated accessibility testing. Pa11y is an automated accessibility testing tool that supports [HTML_Codesniffer](https://squizlabs.github.io/HTML_CodeSniffer/) by Squizlabs, and [axe](https://www.deque.com/axe/) by Deque. 
+
+Projects MUST have Pa11y integrated at the build stage, and builds MUST fail if errors are introduced. You MUST NOT allow accessibility regressions to enter your project's codebase. 
+
```
alexkilgour: > You MUST NOT allow accessibility regressions to enter your project's codebase
Should we mention the reasoning here i.e. The risk of being in breach of contract with regard to our VPATs

accessibility/tools.md
```diff
@@ -0,0 +1,93 @@
...
+You're encouraged to test your pages with assistive technology (AT). Many operating systems include some AT as standard, including screenreaders, magnifiers, or input remapping. 
+
+Make sure you test your work at high magnification. You may also find navigating without an input device using [Voice Control](https://support.apple.com/en-au/HT210539) instructive. Ensure your interfaces work with keyboard navigation. 
+
+Unless you're familiar with using AT, your experience won't be comparable to that of a habitual AT user. Beware of making inferences about what's "easier" for AT users based on your own preferences. 
+
+### Screen readers
+
+If you don’t use a screen reader regularly, the way that you use it will be different to that of a habitual screen reader user, so be cautious in extrapolating your own experiences to those other users. But also bear in mind that not all regular screen reader users are experts or power users, just like not all users of web browsers are experts.
```
alexkilgour: > be cautious in extrapolating your own experiences to those other users

be cautious in extrapolating your own experiences to those OF other users

accessibility/tools.md
```diff
@@ -0,0 +1,93 @@
+# Accessibility tools
+
+- [Automated testing](#automated-testing)
+  - [Pa11y and axe](#pa11y-and-axe)
+- [Manual testing](#manual-testing)
+- [Assistive technology](#assistive-technology)
+  - [Screen readers](#screen-readers)
+- [Other accessibility tools](#other-accessibility-tools)
+  - [Accessibility testing tools in the browser](accessibility-testing-tools-in-the-browser)
+  - [Tools for manually analysing web pages](tools-for-manually-analysing-web-pages)
```
alexkilgour: the links don't work as they are missing the hashes, should be:
- `#accessibility-testing-tools-in-the-browser`
- `#tools-for-manually-analysing-web-pages`

accessibility/tools.md
hollsk: How about

```suggestion
Projects MUST have Pa11y integrated at the build stage, and builds MUST fail if errors are introduced. You MUST NOT allow accessibility regressions to enter your project's codebase. We're aiming to make accessibility better, so our baseline is to not make it worse. 
```

?

alexkilgour: yep, works for me

hollsk:
ready for re-review 😄

accessibility/tools.md
```diff
@@ -0,0 +1,93 @@
...
+
+You're encouraged to test your pages with assistive technology (AT). Many operating systems include some AT as standard, including screenreaders, magnifiers, or input remapping. 
+
+Make sure you test your work at high magnification. You may also find navigating without an input device using [Voice Control](https://support.apple.com/en-au/HT210539) instructive. Ensure your interfaces work with keyboard navigation. 
+
+Unless you're familiar with using AT, your experience won't be comparable to that of a habitual AT user. Beware of making inferences about what's "easier" for AT users based on your own preferences. 
+
+### Screen readers
+
+If you don’t use a screen reader regularly, the way that you use it will be different to that of a habitual screen reader user, so be cautious in extrapolating your own experiences to those of other users. But also bear in mind that not all regular screen reader users are experts or power users, just like not all users of web browsers are experts.
```
jpw: didn't we say this on the previous line?

hollsk: The previous line was about all kinds of AT, not just screen readers. Screen readers have more room for people to believe that the way they're using it is the way everyone else would use it because they're more complicated than other kinds of AT (e.g. there's not too many ways that a person could use a keyboard)

But no other type of AT has a section in this document, so it is a bit redundant. I'll remove the previous line entirely, I think

accessibility/tools.md
```diff
@@ -0,0 +1,91 @@
...
+
+Mature projects using Pa11y for the first time MAY use the `threshold` flag to specify a baseline number of errors to allow through, but MUST aim to reduce that threshold to zero at the earliest opportunity. [Watch this talk by Laura Carvajal of the Financial Times](https://www.youtube.com/watch?v=H4FzW9oFObs) to get a high level overview of the concept. 
+
+Read [Automated accessibility testing with Travis CI](http://cruft.io/posts/automated-accessibility-testing-node-travis-ci-pa11y/) to see how to integrate Pa11y with your build process. If you're not using Travis, adjust the setup for the software that you are using. 
+
+You can use HTML_Codesniffer or axe as your test suite; we recommend using both. Neither engine is "better" than the other, they just use different testing strategies. When you fail builds based on Pa11y results, bear in mind that the two engines return different numbers of results, and that the results are formatted differently. 
+
+When using axe, you may prefer to use the standalone [NPM module](https://github.com/dequelabs/axe-core) directly, instead of wrapping it with Pa11y.
```
dotcode: Is it worth providing a bit of extra info around why someone may want to do this?

dotcode:
This is a great PR Hollie, nice work.

accessibility/tools.md
hollsk: The reason is because of the different results formatting that I mention in the paragraph above. I've just laid it out very badly so the two paragraphs look like they have nothing to do with each other 😂

Good spot! Will do some fiddling so it makes more logical sense.

hollsk: The suggestion to use axe directly was in the original document, which was written back when Pa11y didn't support axe. But now it does, and I can't really think of a justification for having this line in here at all (html_cs doesn't get the same treatment), so I've just removed it. ✨

accessibility/tools.md
```diff
@@ -0,0 +1,89 @@
...
+
+If you’re a Windows user, NVDA is one of the more popular screen readers amongst blind users. This guide from WebAIM [explains the basics of using NVDA to test pages](https://webaim.org/articles/nvda/).
+
+On mobile, VoiceOver is built-in on iOS devices, and TalkBack is usually built-in on Android devices. 71% of respondents to the [WebAIM Screen Reader User Survey #8](https://webaim.org/projects/screenreadersurvey8/#mobilescreenreaders) use iOS VoiceOver on mobile. 
+
+## Other accessibility tools
+
+None of these tools on their own will catch every error. Even by combining all of them, it's still possible to produce an inaccessible webpage. Terrill Thompson [compared several of the most popular tools](http://terrillthompson.com/blog/730) which should give you some idea of the scale of the problem. The Government Digital Service also [conducted their own tool audit](https://accessibility.blog.gov.uk/2017/02/24/what-we-found-when-we-tested-tools-on-the-worlds-least-accessible-webpage/).
```
moddular: This might be fine - the mentions are quite far apart - but it looks like this is the second time we link to the GDS audit?

hollsk: oops, good catch. I combined both of those sections into one, thanks! 😄

PR #349, opened 2020-03-12
add a bit on missing headings

hollsk:
What do when a heading is missing...

### DIFF
--- a/accessibility/common-remediations.md
+++ b/accessibility/common-remediations.md
@@ -809,7 +809,7 @@ A section or context lacks a heading. It may be a section of a page, or a standa
 
 Headings help users understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page. 
 
-If a screen reader user is taken to a new context (e.g. a modal dialog) without a heading, it can be difficult for them to understand what's just happened without tabbing through the rest of the elements to work it out.
+If a screen reader user is taken to a new context (e.g. a modal dialog) without a heading, it can be difficult for them to understand what's just happened without tabbing through the rest of the document to work it out.
 
 #### How do I fix it? 
 
--- a/accessibility/common-remediations.md
+++ b/accessibility/common-remediations.md
@@ -807,7 +807,7 @@ A section or context lacks a heading. It may be a section of a page, or a standa
 
 #### Why's it a problem?
 
-Headings allow users to understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page. 
+Headings help users understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page. 
 
 If a screen reader user is taken to a new context (e.g. a modal dialog) without a heading, it can be difficult for them to understand what's just happened without tabbing through the rest of the elements to work it out.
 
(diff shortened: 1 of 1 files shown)


accessibility/common-remediations.md
```diff
@@ -793,6 +794,29 @@ We _don't_ do this:
   <h3>Headline Topic 2</h3>
 ```
 
+### Missing heading
+
+#### What's the problem?
+
+A section or context lacks a heading. It may be a section on a page, or a standalone context like a modal dialog.
```
jpw:
```suggestion
A section or context lacks a heading. It may be a section of a page, or a standalone context like a modal dialog. 
```

accessibility/common-remediations.md
```diff
@@ -793,6 +794,29 @@ We _don't_ do this:
...
+A section or context lacks a heading. It may be a section on a page, or a standalone context like a modal dialog. 
+
+#### Who's affected by the problem?
+
+* Screen reader users
+* Users with cognitive disabilites or challenges
+
+#### Why's it a problem?
+
+Headings allow users to understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page.
```
jpw:
```suggestion
Headings help users understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page. 
```

accessibility/common-remediations.md
```diff
@@ -793,6 +794,29 @@ We _don't_ do this:
...
+#### Who's affected by the problem?
+
+* Screen reader users
+* Users with cognitive disabilites or challenges
+
+#### Why's it a problem?
+
+Headings allow users to understand the structure of a page. When new sections or contexts are introduced without them, it can be difficult for users to understand how the content relates to the rest of the page. 
+
+If a screen reader user is taken to a new context (e.g. a modal dialog) without a heading, it can be difficult for them to understand what's just happened without tabbing through the rest of the elements to work it out.
```
jpw:
```suggestion
If a screen reader user is taken to a new context (e.g. a modal dialog) without a heading, it can be difficult for them to understand what's just happened without tabbing through the rest of the document to work it out.
```

accessibility/common-remediations.md
```diff
@@ -793,6 +794,29 @@ We _don't_ do this:
...
+
+#### How do I fix it? 
+
+When you change context or open a new section, use a heading to group the content.
```
jpw:
```suggestion
When you change context or open a new section, use a heading to signpost the content. 
```
Unsure about this suggestion. I'm not sure headings "group" content exactly, but I am also not sure how understandable the signpost metaphor is. Enough, maybe?

hollsk: Yeah, I wasn't happy with "group" either as it feels a bit weak as a descriptor. "Signpost" is better but falls foul of the "avoid metaphors" rule, as does "telegraph" or "flag" 😩 

"Announce"? "Signify"? 

Or maybe "open" or "begin the new content"? 🤔

jpw: "Announce" I like, "Signify" (and "demark") more literal but not particularly plain language.

"begin the new content" yep also like. "open", you've just talked about opening a new section, so duplication.

"label" i quite like but would confuse with the HTML element.

I think this is tricky because we're looking for a dual-purpose phrase that needs to relate to "providing a landmark" as this is about navigation, but it's also about describing the content. "Announce" comes close?

I also prefer "begin a new section" I think, rather than "open" in this context, as you have been discussing modals.

"When you change context or begin a new section, use a heading to announce the content." ?

hollsk:
soz I wiped out your approval making that last change 😬happy to give it another tick?

PR #350, opened 2020-03-16
Controls

hollsk:
Happy Monday!

* Renames "links and buttons" section to "controls"
* Adds a bit on things that can go awry with accessible names


accessibility/common-remediations.md
```diff
@@ -195,6 +195,56 @@ We don’t do this:
 <a href="knitting.html" target="_blank">All about Knitting</a>
 ```
 
+### No accessible text on controls
```
alexkilgour: Need to link to this section from the top

hollsk: oops, good spot!

PR #351, opened 2020-03-19
add a bit on use of colour alone

hollsk:
💁 🎨  🖌


hollsk:
You're very kind, thank you! ❤️

PR #356, opened 2020-05-21
Fix wrong first name for quote of Featherstone

morgaan:
It's Derek not Ian

PS: branch name was meant to say `patch` rather than `path`


dotcode:
> It's Derek not Ian

😂

PR #360, opened 2020-06-04
remove recommendation of Lighthouse

hollsk:
We don't recommend the use of this tool anymore. We have other tools that better fit our needs 🙂 

(rationale in this comment if you want the TL;DR: https://github.com/springernature/frontend-playbook/pull/360#issuecomment-638890182 )


saramorueta:
out of curiosity. Do you have a document to read about that conclusion? thanks

hollsk:
We're working on an internal document that does just that! Will let you know once it's ready for publication 😅

dotcode:
Would be interesting to have as much as possible in the public domain …

hollsk:
I've reverted the changes to the performance docs due to overreach 😉

PR #361, opened 2020-06-04
Add line about the motivator for the Playbook

dotcode:


dotcode:
Hmm … interesting 🤔

I _think_ that this particular addition is fine to live in the README, as it's important the motivation is explained up front. In fact I'm horrified it's only now that I've added it, it's 80% the point of the playbook 🤦‍♂️

With regards to the other pages … I think that particular part of the House Style page is almost something you'd expect to see in a Contributing.md. I think this is one of those areas where there's a large degree of overlap between the README/Contributing/house-style pages. That's opening a (small) can of worms, I guess. I think this probably needs looking at, we might be able to figure out a way to reduce the overall number of pages and make navigation easier. Maybe? Worth thinking about, at some point. What do you think?

hollsk:
I opened a new issue for that 😄

hollsk:
I merged this for you because I'm nice like that

PR #366, opened 2020-06-29
Change recommended event system

alexkilgour:
Update the playbook to recommend the new Elements method for passing events between modules. Includes a bit more information on how this differs from our previous recommendation (Pub/Sub).

It may be that we eventually need some form of Pub/Sub, due to the restrictions of DOM binding events, but I think it's better to start simple.

Thoughts welcome!

### DIFF
--- a/javascript/house-style.md
+++ b/javascript/house-style.md
@@ -700,6 +700,7 @@ In the case of two unrelated modules, this would mean that if `moduleA` publishe
 
 ```js
 // Module A
+// Relevant excerpt only
 const event = createEvent('event', 'component', {
 	detail: {
         hazcheeseburger: true
@@ -710,6 +711,7 @@ moduleAelement.dispatchEvent(event);
 
 ```js
 // Module B
+// Relevant excerpt only
 function doSomething(detail) {
 	if (detail.hazcheeseburger) {
 		// Do something
@@ -723,8 +725,10 @@ import {moduleA} from '@springernature/moduleA';
 import {moduleB} from '@springernature/moduleB';
 
 const moduleAelement = document.querySelector('[data-component-module-a]');
+const moduleBelement = document.querySelector('[data-component-module-b]');
 
 moduleA.init(moduleAelement);
+moduleB.init(moduleBelement);
 
 moduleAelement.addEventListener('component:event', function (event) {
 	moduleB.doSomething(event.detail);
--- a/javascript/house-style.md
+++ b/javascript/house-style.md
@@ -199,7 +199,7 @@ Here is an example:
 
 ```js
 ajax('https://example.com', response => {
-  console.log(response)
+    console.log(response)
 });
 ```
 
@@ -207,11 +207,11 @@ If you need to execute another `ajax` request based on the result of the origina
 
 ```js
 ajax('https://example.com', response => {
-  console.log(response); // { status: 200, url: 'https://...' }
+    console.log(response); // { status: 200, url: 'https://...' }
 
-  ajax(response.url, () => {
-    console.log('Two ajax requests have completed')
-  });
+    ajax(response.url, () => {
+        console.log('Two ajax requests have completed')
+    });
 });
 ```
 
@@ -219,9 +219,9 @@ To avoid the journey to '[callback hell](http://callbackhell.com/)', you should 
 
 ```js
 ajax('https://example.com').then(response => response.url).then(url => {
-  ajax(url, () => {
-    console.log('Two ajax requests have completed')
-  });
+    ajax(url, () => {
+        console.log('Two ajax requests have completed')
+    });
 })
 ```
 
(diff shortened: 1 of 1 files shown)


javascript/house-style.md
```diff
@@ -692,20 +692,43 @@ export {init};
 
 #### Events for unrelated modules
 
-Use a small Publish Subscribe implementation, like [PubSubJS](https://github.com/mroderick/PubSubJS):
+Use the [`createEvent` module](https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-javascript#createevent) from the Springer Nature Elements toolkit. This module functions as a namespaced wrapper around Javascript `customEvent`.
 
-```js
-// analytics.js
+One limitation of using the existing infrastructure that Javscript provides — the ability to create an `Event` and consume using `addEventListener` — instead of a [PubSub](https://addyosmani.com/blog/understanding-the-publishsubscribe-pattern-for-greater-javascript-scalability/) style implementation, is that you have to hang the event handler off a `DOM Element` or the `window`.
```
morgaan, line 697:
```suggestion
One limitation of using the existing infrastructure that Javascript provides — the ability to create an `Event` and consume using `addEventListener` — instead of a [PubSub](https://addyosmani.com/blog/understanding-the-publishsubscribe-pattern-for-greater-javascript-scalability/) style implementation, is that you have to hang the event handler off a `DOM Element` or the `window`.
```

javascript/house-style.md
```diff
@@ -692,20 +692,43 @@ export {init};
...
-export {init};
+```js
+// Application
+import {moduleA} from '@springernature/moduleA';
+import {moduleB} from '@springernature/moduleB';
+
+const moduleAelement = document.querySelector('[data-component-module-a]');
+
+moduleA.init(moduleAelement);
+
```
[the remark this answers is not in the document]
alexkilgour: Yes, but I left it out to make things more concise. Maybe it should be in there though.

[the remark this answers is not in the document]
morgaan: IMO, if this saves one line, it does disservice more than it serves, especially if you had workaround by adding `...` as a marker that we are looking at an excerpt of the file and therefore the moduleB init is implicit.
It is really up to you though ;)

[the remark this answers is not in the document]
alexkilgour: No, I think you are right, I will update 😄

javascript/house-style.md
```diff
@@ -692,20 +692,47 @@ export {init};
...
 
-Use a small Publish Subscribe implementation, like [PubSubJS](https://github.com/mroderick/PubSubJS):
+Use the [`createEvent` module](https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-javascript#createevent) from the Springer Nature Elements toolkit. This module functions as a namespaced wrapper around Javascript `customEvent`.
 
-```js
-// analytics.js
+One limitation of using the existing infrastructure that Javascript provides — the ability to create an `Event` and consume using `addEventListener` — instead of a [PubSub](https://addyosmani.com/blog/understanding-the-publishsubscribe-pattern-for-greater-javascript-scalability/) style implementation, is that you have to hang the event handler off a `DOM Element` or the `window`.
 
-import {PubSub} from 'pub-sub';
+In the case of two unrelated modules, this would mean that if `moduleA` publishes an event, then `moduleB` would have no knowledge of the DOM element that the event is bound to. This is solved by having your application manage event listeners, and keeping your modules fully isolated with no knowledge of other modules. A simple example:
```
dotcode, line 699:
```suggestion
In the case of two unrelated modules, this would mean that if `moduleA` publishes an event, then `moduleB` would have no knowledge of the DOM element that the event is bound to. This is solved by having your application manage event listeners, and keeping your modules fully isolated with no knowledge of other modules; for example:
```

javascript/house-style.md
```diff
@@ -692,20 +692,47 @@ export {init};
...
+In the case of two unrelated modules, this would mean that if `moduleA` publishes an event, then `moduleB` would have no knowledge of the DOM element that the event is bound to. This is solved by having your application manage event listeners, and keeping your modules fully isolated with no knowledge of other modules; for example:
 
-function beaconToAnalytics(event) {}
+```js
+// Module A
+// Relevant excerpt only
+const event = createEvent('event', 'component', {
+	detail: {
+        hazcheeseburger: true
+    }
```
danyalaytekin, line 707:
```suggestion
    detail: {
        hazcheeseburger: true
    }
```

Also the code samples in this document in general use a mixture of tabs, 2-char soft tabs and 4-char soft tabs.  Was that intentional, with the concerns in the samples being independent of indentation concerns, or is it an oversight?

alexkilgour: Hey @danyalaytekin, sorry been on holiday. Thanks for the review 😄  This mixture of spacing was not intentional. I will go through and standardise.

PR #368, opened 2020-09-21
rejig the package installation instructions

hollsk:
* Move `npm ci` up in priority
* Move `npm install` down in priority
* Add a bit about when you might not want to use `npm ci`
* Removed the bit about specifying a newer version of npm than is recommended for your version of node (i'm amenable to keeping it if there are objections to this - I dropped it in the name of simplicity)

### DIFF
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -40,7 +40,7 @@ It is the method recommended for installing dependencies when authoring node _ap
 
 ### Using the right version of node
 
-When authoring libraries you are probably supporting multiple versions of node, so it not required to use a specific version when installing (as long as the version is supported by your library).
+When authoring libraries you're probably supporting multiple versions of node, so you're not required to use a specific version when installing (as long as the version is supported by your library).
 
 When authoring node applications, and therefore using `npm ci`, it is important to specify a version of node to use. This is because the `package-lock.json` file will reflect the version of node used when the `package-lock.json` file was created.  
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -48,9 +48,9 @@ If your project has a `package-lock.json` file (see [Should I check in `package-
 
 #### nvm users run `nvm use` before `npm install`
 
-If `npm ci` is unavailable (because you're using an older version of npm), or otherwise unsuitable (e.g. you have no `package-lock.json`, or you're installing new/updating existing dependencies), use `npm install`. 
-
 If you're an nvm user, you should always ensure you're using the correct version of node (and implicitly `npm`) before doing an `npm install`.
+
+If `npm ci` is unavailable (because you're using an older version of npm), or otherwise unsuitable (e.g. you have no `package-lock.json`, or you're installing new/updating existing dependencies), use `npm install`. 
 
 Using `npm install` doesn't guarantee reproducible builds, as neither the `package.json` nor `package-lock.json` is a source of authority for what's installed. 
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -26,7 +26,7 @@ The following guide describes how we use node and manage package dependencies in
 
 There are two methods for installing node dependencies: `npm install` and `npm ci`.
 
-`npm install` is the most common method, but it does not guarantee reproducible/consistent builds because npm can install varying versions of dependencies, according to [semantic versioning](https://docs.npmjs.com/files/package.json#dependencies) and the version of node used when installing. When running `npm install`, it will create a `package-lock.json` file describing the complete dependency graph and specifying exact versions of all the dependencies installed.
+`npm install` is the most common method, but it doesn't guarantee reproducible/consistent builds. npm can install varying versions of dependencies, according to [semantic versioning](https://docs.npmjs.com/files/package.json#dependencies) and the version of node used when installing. `npm install` will create a `package-lock.json` file describing the complete dependency graph and specifying exact versions of all the dependencies installed.
 
 Note that if a `package-lock.json` is present **`npm install` will not use it** as the source of authority for what to install.
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -85,7 +85,7 @@ The downside is it requires all developers to use `nvm` and `npm ci` to install 
 
 Firstly, bear in mind machine-generated files should not be hand-edited, and `package-lock.json` is no exception.
 
-Secondly, if you are seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
+Secondly, if you're seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
 
 When you update your application, we strongly recommend that you keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -78,7 +78,7 @@ which run `nvm` and switch node versions for you when `cd`-ing into a directory.
 Yes. There are two main reasons:
 
 1. Predictable builds, as discussed above.
-1. Dependency analysis tools (such as `npm audit`, GitHub Security Alerts and `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. Furthermore these tools can force patch version updates of insecure dependencies way down the tree, rather than having to wait for all the package maintainers down that branch of the tree to release new packages with updated dependencies. This would not be possible without a `package-lock.json` and is a huge benefit.
+1. Dependency analysis tools (such as `npm audit`, GitHub Security Alerts and `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. These tools can force patch version updates of insecure dependencies in the tree, without waiting for third-party package maintainers to release new packages with updated dependencies. This wouldn't be possible without a `package-lock.json` and is a huge benefit.
 
 The downside is it requires all developers to use `nvm` and `npm ci` to install dependencies, otherwise there will be constant conflicts in the `package-lock.json` file.
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -1,6 +1,8 @@
 # Managing Node.js-based projects and dependencies
 
 - [Installing dependencies](#installing-dependencies)
+  - [`npm install`](#npm-install)
+  - [`npm ci`](#npm-ci)
   - [Using the right version of node](#using-the-right-version-of-node)
   - [How to specify the version of node to use](#how-to-specify-the-version-of-node-to-use)
     - [`nvm` usage](#nvm-usage)
@@ -26,11 +28,15 @@ The following guide describes how we use node and manage package dependencies in
 
 There are two methods for installing node dependencies: `npm install` and `npm ci`.
 
+### `npm install`
+
 `npm install` is the most common method, but it doesn't guarantee reproducible/consistent builds. npm can install varying versions of dependencies, according to [semantic versioning](https://docs.npmjs.com/files/package.json#dependencies) and the version of node used when installing. `npm install` will create a `package-lock.json` file describing the complete dependency graph and specifying exact versions of all the dependencies installed.
 
 Note that if a `package-lock.json` is present **`npm install` will not use it** as the source of authority for what to install.
 
 This is the method we recommend for installing dependencies when authoring node _libraries_.
+
+### `npm ci`
 
 `npm ci` installs the complete dependency graph exactly as specified in the `package-lock.json` file, and so guarantees reproducible/consistent builds. This helps reduce "but it works on my machine" issues and is suitable for CI/CD environments (as the name implies).
 
@@ -42,7 +48,7 @@ This is the method we recommend for installing dependencies when authoring node 
 
 When authoring libraries you're probably supporting multiple versions of node, so you're not required to use a specific version when installing (as long as the version is supported by your library).
 
-When authoring node applications, and therefore using `npm ci`, it is important to specify a version of node to use. This is because the `package-lock.json` file will reflect the version of node used when the `package-lock.json` file was created.  
+When authoring node applications, and therefore using `npm ci`, it is important to specify a version of node to use. This is because the `package-lock.json` file will reflect the version of node used when the `package-lock.json` file was created.
 
 If you don't use the same version of node when running `npm ci`, the versions of dependencies specified in the `package-lock.json` may be incompatible with the version of node you are currently using, and you may get installation errors.
 
--- a/practices/managing-node-projects.md
+++ b/practices/managing-node-projects.md
@@ -1,17 +1,19 @@
 # Managing Node.js-based projects and dependencies
 
-- [Specifying versions of node](#specifying-versions-of-node)
-  - [Installing packages predictably](#installing-packages-predictably)
-    - [Prefer `npm ci` to `npm install` where possible](#prefer-npm-ci-to-npm-install-where-possible)
-    - [nvm users run `nvm use` before `npm install`](#nvm-users-run-nvm-use-before-npm-install)
-      - [Automatically running `nvm use`](#automatically-running-nvm-use)
+- [Installing dependencies](#installing-dependencies)
+  - [Using the right version of node](#using-the-right-version-of-node)
+  - [How to specify the version of node to use](#how-to-specify-the-version-of-node-to-use)
+    - [`nvm` usage](#nvm-usage)
+    - [Automatically running `nvm use`](#automatically-running-nvm-use)
   - [Should I check in `package-lock.json` to version control?](#should-i-check-in-package-lockjson-to-version-control)
+  - [Managing changes to the `package-lock.json` file](#managing-changes-to-the-package-lockjson-file)
 - [Specifying versions of dependencies](#specifying-versions-of-dependencies)
   - [Run-time dependencies](#run-time-dependencies)
     - [Examples](#examples)
   - [Development dependencies](#development-dependencies)
     - [Examples](#examples-1)
   - [Classifying "built" run-time dependencies](#classifying-built-run-time-dependencies)
+- [Configuration](#configuration)
 - [Publishing projects on NPM](#publishing-projects-on-npm)
   - [Naming projects](#naming-projects)
 - [Dependency management tools](#dependency-management-tools)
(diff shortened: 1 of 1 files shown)


practices/managing-node-projects.md
```diff
@@ -34,23 +36,25 @@ It's important to specify which versions of node your application expects. There
...
 
-[`npm ci` is also much quicker than `npm install`](https://docs.npmjs.com/cli/ci.html#description) if the `./node_modules` directory is not present (such as in a CI environment).
+#### nvm users run `nvm use` before `npm install`
 
-(You could use `npm ci` by specifying a newer version of `npm` than is recommended for your [particular version of node](https://nodejs.org/en/download/releases/), but `npm` is itself written in node and [only supports certain node versions](https://github.com/npm/cli/blob/latest/lib/utils/unsupported.js).)
+If `npm ci` is unavailable (because you're using an older version of npm), or otherwise unsuitable (e.g. you have no `package-lock.json`, or you're installing new/updating existing dependencies), use `npm install`.
```
benjclark, line 51: image
 
This paragraph about falling back to `npm install` looks sandwiched inbetween the two paragraphs that should be next to each other that describe how to ensure the correct node version.

Perhaps move the paragraph down so it's just above the paragraph beginning with "Using npm install does't guarantee reproducible builds..."?

hollsk: oh yeah! good shout

dotcode:
What lovely clarity. Great stuff.

jpw:
This is much improved Hollie, thank you.

We should probably add what @josebolos is probably going to mention, that this applies to node _applications_ not libraries. For libraries, using a package lock is much more debatable and probably inadvisable. Not sure if that should wait for another PR.

hollsk:
@jpw I added a little bit to the start with that in it, but what I'd really like is to link to something coherent that explains the debate and the inadvisability. I don't suppose any of you have such a URL we can sling in there?

jpw:
https://docs.npmjs.com/files/package-lock.json

> One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package. It shares a format with npm-shrinkwrap.json, which is essentially the same file, but allows publication. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.

https://docs.npmjs.com/files/shrinkwrap.json#description

> The recommended use-case for npm-shrinkwrap.json is applications deployed through the publishing process on the registry: for example, daemons and command-line tools intended as global installs or devDependencies. It’s strongly discouraged for library authors to publish this file, since that would prevent end users from having control over transitive dependency updates.

This means that
1) The `package-lock.json` cannot be published to npm, but libs (not apps) typically _are_ published to npm. If lib authors used a `package-lock.json` _locally_ (despite the fact it cannot be published), they would not have the same experience as users of the library. One reason this is bad is that they would be less likely to have their own builds breaking when their consumers builds are breaking -- for example when new versions of the libraries dependencies are published, and maintainer of such a dependency has a semver accident and breaks some consumers of their package (sadly not that rare an occurence). You as a lib author would not experience this if you have your lib's dependencies locked locally, but your users would.
2) `npm-shrinkwrap.json` is the same as a `package-lock.json`, but publishable to npm in the event you really _really_ want to do that, for some weird reason.  If you _did_ use an `npm-shrinkwrap.json` for your library, this would prevent users of your lib using tools like `npm audit fix` to fix deps of deps, because your lockfile has locked all the deps. This means users could not use various tools to patch your insecure lib. So they would have to do this manually (a PITA is this is a generated file) or wait for you to publish a new version containing securer versions of your deps.

hollsk:
Hrm. Do you want me to rewrite that and put it in? I was hoping for a link to an external resource so I didn't have to do that! 

From the [Technical writing](https://github.com/springernature/frontend-playbook/blob/master/writing/house-style.md#technical-writing) bit in the house style:
>Your priorities must be clarity, brevity, and accuracy. If you think your reader may need additional context to understand a concept, link out to external resources.

jpw:
I think a rewrite, yes please. I looked at a few docs on the topic (from some reputable sources) and none were particularly clear -- it is quite a tricky topic to get ones head around. The docs on npmjs.com are obviously a Source of Authority, but the implications of their docs are naturally not included.

josebolos:
I would suggest to separate this work with the work related to the package-lock file in order to not make the PR unmergeable.

practices/managing-node-projects.md
```diff
@@ -34,23 +36,25 @@ It's important to specify which versions of node your application expects. There
...
-You should always ensure you're using the correct version of node (and implicitly `npm`) before doing an `npm install`.
+To reduce "but it works on my machine" issues, use a method that makes installs more predictable, and minimises changes to the `package-lock.json` (if your project is committing that file - this will be the case if your software is a node _application_, but not if it's a _library_).
 
-Running `nvm use` before `npm install` is good because:
-1. It make installs _more_ predictable, which minimises "but it works on my machine" issues.
-1. It minimises changes to the `package-lock.json` (if your project is comitting that file).
+#### Prefer `npm ci` to `npm install` where possible
 
-A problem with using `npm install` is that it doesn't guarantee reproducible builds, as neither the `package.json` nor `package-lock.json` is a source of authority for what's installed.
+The `npm ci` command was introduced with `npm` version 5.7.0+ ([node v10.3+ or 8.12+](https://nodejs.org/en/download/releases/)).
```
josebolos, line 45: For simplicity, maybe we could remove this altogether. We shouldn't be using a version of Node.js that bundles a version of npm that doesn't support `npm ci`, so removing this would make the document less scary and easier to understand.

practices/managing-node-projects.md
```diff
@@ -34,23 +36,25 @@ It's important to specify which versions of node your application expects. There
...
 
-[`npm ci` is also much quicker than `npm install`](https://docs.npmjs.com/cli/ci.html#description) if the `./node_modules` directory is not present (such as in a CI environment).
+#### nvm users run `nvm use` before `npm install`
 
-(You could use `npm ci` by specifying a newer version of `npm` than is recommended for your [particular version of node](https://nodejs.org/en/download/releases/), but `npm` is itself written in node and [only supports certain node versions](https://github.com/npm/cli/blob/latest/lib/utils/unsupported.js).)
+If you're an nvm user, you should always ensure you're using the correct version of node (and implicitly `npm`) before doing an `npm install`.
 
-#### Automatically running `nvm use`
+If `npm ci` is unavailable (because you're using an older version of npm), or otherwise unsuitable (e.g. you have no `package-lock.json`, or you're installing new/updating existing dependencies), use `npm install`.
```
josebolos, line 53: Same as above, I would change this so it suggests:

* `npm ci` for installing "prod ready" dependencies
* `npm install` to install new dependencies or bump transitive deps

practices/managing-node-projects.md
jpw: Agreed

practices/managing-node-projects.md
jpw: Shall we get this PR merged and consider this later as well?

jpw:
Hello, I had a rush of blood to the head and decided to rewrite a big chunk of this document. Sorry for invalidating some of the PR chat this far.

I had the feeling we were struggling with this document because it was not particularly logically-organised in the first place (my fault). I switched the order around a bit, to;

1. Installing dependencies (`npm i` vs `npm ci` -- which originally was not a concept introduced first, which makes no sense)
1. Using the right version of node
1. How to specify the version of node to use
1. nvm usage
1. Should I check in package-lock.json to version control?
1. Managing changes to the package-lock.json file (new)

..which I hope introduces the concepts in a more logical order. Previously it started talking about npm install, then nvmrc, then using nvmrc, then npm ci, which was confusing and reasoning was scattered around in the document.

I also call out the differences when authoring apps vs libs as I go, as again this relates to when to use `npm i` vs `npm ci`

practices/managing-node-projects.md
```diff
@@ -18,54 +22,74 @@
...
+2. Using an `.nvmrc` file:
+     * An `.nvmrc` file is a configuration file for `nvm` ([Node Version Manager](https://github.com/creationix/nvm)).
+     Use of `nvm` is strongly encouraged as an application author, as it enables you to specify to other developers what versions of node they should be using when running your app.
+     * Your application should include an [`.nvmrc` file](https://github.com/creationix/nvm#nvmrc) in the root directory of the project to specify which version(s) of node are compatible.
+     * It also is required for proper `npm ci` usage, as explained above.
+     * It also makes life much easier when working on multiple node projects locally that require different versions of node.
+
+#### `nvm` usage
+
+When developers check out your project, they should first run `nvm use` before `npm ci` to ensure the project is running on the same version of node as you. Additionally [Travis respects `.nvmrc` files](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc), so using one will simplify your Travis configuration.
```
hollsk, line 66:
```suggestion
When you check out a project, you should first run `nvm use` before `npm ci` to ensure you're running on the same version of node the project expects. Additionally [Travis respects `.nvmrc` files](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc), so using one will simplify your Travis configuration and help other developers avoid mistakes.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +22,74 @@
...
 
-Yes, if you're using a version of `npm` that supports the `ci` argument. There are two main reasons:
+Yes. There are two main reasons:
 
 1. Predictable builds, as discussed above.
-1. Dependency analysis tools (such as `npm audit` & `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. Furthermore these tools can force patch version updates of insecure dependencies way down the tree, rather than having to wait for all the package maintainers down that branch of the tree to release new packages with updated dependencies. This would not be possible without a `package-lock.json`.
+1. Dependency analysis tools (such as `npm audit`, GitHub Security Alerts and `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. Furthermore these tools can force patch version updates of insecure dependencies way down the tree, rather than having to wait for all the package maintainers down that branch of the tree to release new packages with updated dependencies. This would not be possible without a `package-lock.json` and is a huge benefit.
```
hollsk, line 81:
```suggestion
1. Dependency analysis tools (such as `npm audit`, GitHub Security Alerts and `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. These tools can force patch version updates of insecure dependencies in the tree, without waiting for third-party package maintainers to release new packages with updated dependencies. This wouldn't be possible without a `package-lock.json` and is a huge benefit.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +22,74 @@
...
 
 The downside is it requires all developers to use `nvm` and `npm ci` to install dependencies, otherwise there will be constant conflicts in the `package-lock.json` file.
 
+### Managing changes to the `package-lock.json` file
+
+Firstly, bear in mind machine-generated files should not be hand-edited, and `package-lock.json` is no exception.
+
+Secondly, if you are seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
+
+Additionally, when updating your application its strongly recommend to keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` to their own PR, free of other changes to the application. This helps by making your PR easier to review.
```
hollsk, line 91:
```suggestion
Additionally, when updating your application its strongly recommend to keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
```

hollsk (commented):
Use or chuck suggestions as your fancy takes 'em 💁‍♂️

hollsk:
Looks good to me, and thank you, but I can't approve it because it's my pull request 😂 I think you'll need to actively seek input from the others 💖

jpw:
Aha, good point. What do you think @josebolos ? Do you think this PR;

 - is generally more readable
 - now properly distinguishes between app & library use-cases?

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
...
+
+1. The [`engines`](https://docs.npmjs.com/files/package.json#engines) field in `package.json`:
+     * `engines` is important if authoring libraries. Let's assume we're using a different version of node to that specified in `engines` field for package `foo`.
+     When running `npm install` to install `foo`'s dependencies, `npm` will warn about the problem.
+     When `npm install`ing a package _which depends on_ `foo`, `npm` will warn and error.
+     * Some deployment environments also respect it.
+     * It makes compatibility requirements explicit to developers working on your application.
+2. Using an `.nvmrc` file:
+     * An `.nvmrc` file is a configuration file for `nvm` ([Node Version Manager](https://github.com/creationix/nvm)).
+     Use of `nvm` is strongly encouraged as an application author, as it enables you to specify to other developers what versions of node they should be using when running your app.
```
josebolos, line 67: > Use of `nvm` is strongly encouraged as an application author

We were talking about "applications" vs "libraries" before, the use of "application" here is confusing I think.

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
...
+     * Your application should include an [`.nvmrc` file](https://github.com/creationix/nvm#nvmrc) in the root directory of the project to specify which version(s) of node are compatible.
+     * It also is required for proper `npm ci` usage, as explained above.
+     * It also makes life much easier when working on multiple node projects locally that require different versions of node.
+
+#### `nvm` usage
+
+When you check out a project, you should first run `nvm use` before `npm ci` to ensure you're running on the same version of node the project expects. Additionally [Travis respects `.nvmrc` files](https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions-using-nvmrc), so using one will simplify your Travis configuration and help other developers avoid mistakes.
 
 #### Automatically running `nvm use`
```
josebolos, line 76: I don't think this needs to be a separate point to "nvm usage"

josebolos (commented):
This is great but I'm not sure it's getting the point across. I think that it needs a simplified structure. I added a few comments about specific issues.

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
...
 
 The downside is it requires all developers to use `nvm` and `npm ci` to install dependencies, otherwise there will be constant conflicts in the `package-lock.json` file.
 
+### Managing changes to the `package-lock.json` file
+
+Firstly, bear in mind machine-generated files should not be hand-edited, and `package-lock.json` is no exception.
+
+Secondly, if you are seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
+
+Additionally, when updating your application its strongly recommend to keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
```
danyalaytekin, line 99:
```suggestion
Additionally, when updating your application it's strongly recommend to keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
```

practices/managing-node-projects.md
jpw: I am still talking about applications here.

practices/managing-node-projects.md
jpw: Sorry, can you rephrase that please?

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
 The following guide describes how we use node and manage package dependencies in [Springer Nature Node.js-based projects](https://github.com/springernature?utf8=%E2%9C%93&q=&type=public&language=javascript).
 
 
-## Specifying versions of node
+## Installing dependencies
 
-It's important to specify which versions of node your application expects. There are two ways of doing this, and you're encouraged to do both.
+There are two methods for installing node dependencies: `npm install` and `npm ci`.
```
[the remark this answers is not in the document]
jpw: I disagree with your approach as there are several use cases;
 - "I just want to install the things" vs "i am seeking to actively develop this project"
 - apps vs libs
 - projects may or may not properly have a lockfile checked in

There are too many permutations to produce a prescriptive guide, and in any case I do not recall any principles of the runbook saying that we default to that approach.

"How to install node deps" seems a good place to start to teach these topics, rather than starting to talk about lockfiles, when lockfiles may or may not be relevant to my use case, and I may not even know what a lockfile is. For these reasons this doc is intended to teach concepts around `i` & `ci` and is more "educational" than most -- it is a complicated topic.

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
...
 [`npm ci` is also much quicker than `npm install`](https://docs.npmjs.com/cli/ci.html#description) if the `./node_modules` directory is not present (such as in a CI environment).
 
-(You could use `npm ci` by specifying a newer version of `npm` than is recommended for your [particular version of node](https://nodejs.org/en/download/releases/), but `npm` is itself written in node and [only supports certain node versions](https://github.com/npm/cli/blob/latest/lib/utils/unsupported.js).)
+This is the method we recommend for installing dependencies when authoring node _applications_.
+
+### Using the right version of node
+
+When authoring libraries you're probably supporting multiple versions of node, so you're not required to use a specific version when installing (as long as the version is supported by your library).
+
+When authoring node applications, and therefore using `npm ci`, it is important to specify a version of node to use. This is because the `package-lock.json` file will reflect the version of node used when the `package-lock.json` file was created.
```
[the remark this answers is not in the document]
jpw: I think it is true but am having a hard time finding anything beyond anecdata and random SO posts, so it may well be a superstition :)

I'm sure I remember problems with node-sass in this area, and fsevents, but I think these are examples where it _will or wont install_ according to _platform_. It's also true of course that packages may or may not install due to the user not using a supported version of node. But these are examples of a package either installing or not, and not examples of a successful install but of different _versions_ of the package.

Another thing to support that theory is that I cannot see a way to target different versions of packages at different versions of node.

So I think I should amend this line, and the following line, thanks!

practices/managing-node-projects.md
```diff
@@ -18,54 +24,80 @@
...
+If you don't use the same version of node when running `npm ci`, the versions of dependencies specified in the `package-lock.json` may be incompatible with the version of node you are currently using, and you may get installation errors.
+
+### How to specify the version of node to use
+
+There are two main ways of doing this, and you're encouraged to do both.
+
+1. The [`engines`](https://docs.npmjs.com/files/package.json#engines) field in `package.json`:
+     * `engines` is important if authoring libraries. Let's assume we're using a different version of node to that specified in `engines` field for package `foo`.
+     When running `npm install` to install `foo`'s dependencies, `npm` will warn about the problem.
+     When `npm install`ing a package _which depends on_ `foo`, `npm` will warn and error.
```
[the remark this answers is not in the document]
jpw: Yep, will amend, thanks!

practices/managing-node-projects.md
jpw: oh, i see what you mean. IIRC Hollie disagrees :) I think I do too. Using nvm is one topic, automatically using it is a different point.

practices/managing-node-projects.md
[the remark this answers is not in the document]
jpw: Have removed this section as I think its pointless.

practices/managing-node-projects.md
```diff
@@ -1,15 +1,21 @@
 # Managing Node.js-based projects and dependencies
```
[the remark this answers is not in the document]
jpw: Packages are listed as "dependencies" or "devDependencies" in the `package. json` and we start discussing the installation of these dependency packages straight away, so to me it makes more sense to use "dependencies". Also I think developers are more likely to have looked at a `package.json` than the npm docs.

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
 
 #### Automatically running `nvm use`
 
-To avoid accidentally forgetting to run `nvm use` before working on a node project, there are useful shell extensions which will do it for you
-([zsh](https://github.com/creationix/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file),
-[bash](https://stackoverflow.com/questions/23556330/run-nvm-use-automatically-every-time-theres-a-nvmrc-file-on-the-directory)),
-which are well worth installing!
+As mentioned above, it is important to run `nvm use` before `npm ci`. But this is a thing that's easy to forget!
+
+To avoid forgetting, there are useful shell extensions ([zsh](https://github.com/creationix/nvm#calling-nvm-use-automatically-in-a-directory-with-a-nvmrc-file),
```
springerBuck, line 71: would it be useful to suggest that a `preinstall` can be used to run `nvm use` before anything is installed? it wouldn't help with changing between projects, but it would help when installing modules for the first time and could be a useful defensive alternative that a developer could implement

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
 
-(You could use `npm ci` by specifying a newer version of `npm` than is recommended for your [particular version of node](https://nodejs.org/en/download/releases/), but `npm` is itself written in node and [only supports certain node versions](https://github.com/npm/cli/blob/latest/lib/utils/unsupported.js).)
+This is the method we recommend for installing dependencies when authoring node _applications_.
+
+### How to specify the version of node to use
+
+There are two main ways of doing this, and you're encouraged to do both.
+
+1. The [`engines`](https://docs.npmjs.com/files/package.json#engines) field in `package.json`:
+     * `engines` is important if authoring libraries. Let's assume we're using a different version of node to that specified in `engines` field for package `foo`.
```
hollsk, line 51:
```suggestion
     * `engines` is important if you're authoring libraries. Let's assume we're using a different version of node to that specified in `engines` field for package `foo`.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
+     When running `npm install` to install `foo`'s dependencies, `npm` will warn about the problem.
```
hollsk, line 52:
```suggestion
     When you run `npm install` to install `foo`'s dependencies, `npm` will warn about the problem.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
+     When `npm install`ing a package _which depends on_ `foo`, `npm` will warn.
+     * Some deployment environments also respect it.
+     * It makes compatibility requirements explicit to developers working on your application.
+2. Using an `.nvmrc` file:
+     * An `.nvmrc` file is a configuration file for `nvm` ([Node Version Manager](https://github.com/creationix/nvm)).
+     Use of `nvm` is strongly encouraged as an application author, as it enables you to specify to other developers what versions of node they should be using when running your app.
```
hollsk, line 58:
```suggestion
     We strongly encourage you to use `nvm` when you're authoring applications, as it enables you to specify to other developers what versions of node they should be using when running your app.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
+1. Dependency analysis tools (such as `npm audit`, GitHub Security Alerts and `snyk`) can spot insecure dependencies anywhere in the dependency tree by analysing the `package-lock.json` file. These tools can force patch version updates of insecure dependencies in the tree, without waiting for third-party package maintainers to release new packages with updated dependencies. This wouldn't be possible without a `package-lock.json` and is a huge benefit.
 
 The downside is it requires all developers to use `nvm` and `npm ci` to install dependencies, otherwise there will be constant conflicts in the `package-lock.json` file.
 
+### Managing changes to the `package-lock.json` file
+
+Firstly, bear in mind machine-generated files should not be hand-edited, and `package-lock.json` is no exception.
+
+Secondly, if you are seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
```
hollsk, line 88:
```suggestion
Secondly, if you're seeing unexpected merge conflicts in `package-lock.json` it can be a symptom of someone using `npm install` instead of `npm ci`. Speak to the developer and see if they are having trouble.
```

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
+
+Additionally, when updating your application it's strongly recommended to keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
```
hollsk, line 90:
```suggestion
When you update your application, we strongly recommend that you keep `package-lock.json` changes as a result of `npm update` or `npm audit fix` in a separate PR to other changes to the application. This helps by making your PRs easier to review.
```

hollsk (commented):
👋

practices/managing-node-projects.md
jpw: Hmm, interesting point, thanks 🙂

I do get your argument, but in _this_ case I think it's probably better to crash early with errors on install TBH, to make developers aware of the issue. Anecdata but, I don't think I've ever used `preinstall` so this bit of magic might come as a bit of a surprise?

And yes as you say, it won't help with switching between projects (or the global node & npm versions being hiked). So am inclined to keep it simple...

practices/managing-node-projects.md
```diff
@@ -18,54 +23,72 @@
...
+     * `engines` is important if you're authoring libraries. Let's assume we're using a different version of node to that specified in `engines` field for package `foo`.
+     When you run `npm install` to install `foo`'s dependencies, `npm` will warn about the problem.
+     When `npm install`ing a package _which depends on_ `foo`, `npm` will warn.
+     * Some deployment environments also respect it.
+     * It makes compatibility requirements explicit to developers working on your application.
+2. Using an `.nvmrc` file:
+     * An `.nvmrc` file is a configuration file for `nvm` ([Node Version Manager](https://github.com/creationix/nvm)).
+     We strongly encourage you to use `nvm` when you're authoring applications, as it enables you to specify to other developers what versions of node they should be using when running your app.
+     * Your application should include an [`.nvmrc` file](https://github.com/creationix/nvm#nvmrc) in the root directory of the project to specify which version(s) of node are compatible.
+     * It also is required for proper `npm ci` usage, as explained above.
```
jon-stevens, line 60: Should this be 'explained below'?

jpw: Nice spot thanks, fixed!

practices/managing-node-projects.md
springerBuck: Yep, that seems a fair point. Better to keep things simple

jon-stevens:
It looks pretty darn excellent to me

jon-stevens (approved):

jpw:
Thanks everyone for your thoughts on this large PR. It's been a journey 🎸

PR #370, opened 2020-09-25
Link to our own caching documentation

josebolos:


efasel:
Would this also be relevant/interesting in this context?
[Caching Guide - Engineering Enablement](https://ee.public.springernature.app/cdn/)

josebolos:
> Would this also be relevant/interesting in this context?
> [Caching Guide - Engineering Enablement](https://ee.public.springernature.app/cdn/)

Thanks @efasel! I wasn't aware of that link. We like to keep the playbook simple, but we may be able to add a link from the caching page that I'm linking to, to the EE one.

jon-stevens:
👍

dotcode (approved):

PR #371, opened 2020-10-06
Update introduction.md

emsel5201:
Remove the Financial Times link as now a dead-link.

### DIFF
--- a/accessibility/introduction.md
+++ b/accessibility/introduction.md
@@ -100,8 +100,6 @@ You can arrange your own user testing, but it's better to engage a third party l
 
 Udacity have published an [excellent free course on Web Accessibility](https://www.udacity.com/course/web-accessibility--ud891) with Google engineers Alice Boxhall and Rob Dodson. 
 
-The Financial Times have produced an [accessibility tip sheet](https://ft-interactive.github.io/accessibility/index.html).
-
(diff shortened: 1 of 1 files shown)


josebolos (approved):
Good catch! 👌

sonniesedge:
Nice spot. The repo that generated the page is still present, but has been archived, which would explain the Github Pages URL dying.
https://github.com/ft-interactive/accessibility

Should we ask if they can revive it?

sonniesedge:
(I have forked it onto my own account in case they decide to kill the repo as well)

hollsk:
Not sure how/who we'd ask, to be honest. We can't open an issue on an archived repo, and I don't recognise either of the contributors 😞 Do you know them?

sonniesedge:
I can ping one of them for sure. They might not even be aware the site is down. Will update here if I hear back from them.

sonniesedge:
Good news and bad news! 

🎉 Good: they had deleted that page in error and have now restored it. (thanks for restoring it @joannaskao!)
👿 Bad: someones will need to restore the link on our Playbook. 😂

hollsk:
You're a star, Charlie, thank you. I don't know what we've done to deserve you 🤗

Issue #135, opened 2017-02-16
Document our approach(es) to webfont loading


hollsk:
Summary of our discussion earlier today:

> I think we were agreed on loading them with JS.  There was discussion about non js users not getting the font, but that's ok I think.  Even with the JS approach, there's differences between projects.  We use session storage to set a flag once font has loaded, but link save the entire font as a data-uri in local storage (I think?)

>we do save the entire font as a data-uri then retrieve it from local storage. I actually went through that article that was posted and tested out various options and we found this one to be the best for us. BUT it might not be best for all. We only have one webfont, and it has a limited character set so storing at as a data-uri works fine. not sure how well it would work with multiple fonts

>the way we implement it is slightly different from how it is described in the article (https://www.zachleat.com/web/comprehensive-webfonts/#critical-foft-data-uri) as we don't write it inline

>I actually think it is OK that we use slightly different methods, there is still a lot of common ground and we can probably write up a set of broader guidelines and then detail the individual approaches. 

>not like a document that decrees YOU MUST DO XYZ but more a document that outlines the problems, what we tried, and some recommended options

danyalaytekin:
Are webfonts one place where we can actually say, within Springer Nature at least, that we could achieve a single approach, and that "we should do XYZ”?  

Some other things we do, like our varied CSS approaches, require less prescriptive documentation but maybe our webfonts approach belongs more with commandments like our CTM method and “don’t self-close tags”.

joseluisbolos:
I was wondering about this too. If there's a "best approach" that we can agree on and solves most problems it would be better to try to stick to it, rather than "do whatever you want".

jpw:
Consistency FTW. If some universal best practice does emerge, it's easier to fix in one place than many.

nickcall:
Alex will submit a PR

Sreilys:
What do you think about following this article, meantime we set a "standard" way? 
https://www.html5rocks.com/en/tutorials/webfonts/quick/
What do you think about using WebFont Loader approach?

davidpauljunior:
More investigation needed I think to agree on a "standard" approach.

sonniesedge:
Overview of font projects:

## Basic loaders and font-loading monitors:

- https://github.com/bramstein/fontfaceobserver
- https://github.com/zachleat/fontfaceonload

Both fontfaceobserver and fontfaceonload rely on the dev loading the fonts. They listen for the font being loaded and then make a callback available, where you can do what you want. 

[...]

PR #378, opened 2021-04-13
Updates guidance on performance - font loading

benjclark:
As per our recent discussion in a frontend open space with regards to font loading: https://springernature.slack.com/archives/C1HGRAQ8G/p1617887140107800

This updates the playbook with font loading guidance. It does not prescribe a particular technique, but provides helpful information and areas to consider.

Also, once this is merged we can close https://github.com/springernature/frontend-playbook/issues/135


performance/font-loading.md
```diff
@@ -0,0 +1,56 @@
+# Font Loading
+
+## Are you sure you need to use webfonts?
+
+It's worth asking this question initially. What benefit do you believe a webfont will give you and is it worth the cost in page performance? Afterall, system fonts are battle tested and were produced with legibility in mind.
+
+## Implementation
+We do not prescribe a particular implementation. Each team should  determine which implementation is best suited for them. However, before selecting an implementation a team should be mindful of our working practises such as:
```
springerBuck, line 8:
```suggestion
We do not prescribe a particular implementation. Each team should determine which implementation is best suited for them. However, before selecting an implementation a team should be mindful of our working practices such as:
```
spotted a typo and double space :)

benjclark: Thank you!

performance/font-loading.md
```diff
@@ -0,0 +1,56 @@
+# Font Loading
+
+## Are you sure you need to use webfonts?
+
+It's worth asking this question initially. What benefit do you believe a webfont will give you and is it worth the cost in page performance? Afterall, system fonts are battle tested and were produced with legibility in mind.
```
josebolos, line 5: I love the fact that this is the first item. I would suggest rephrasing it so the title is not a question, to something like:


> ## When to use webfonts
> 
> Before using webfonts it's worth considering what benefits webfonts provide, and what impact they may have on things like accessibility or page speed. For example:
> 
> * System fonts are battle tested and were produced with legibility in mind. Using webfonts may reduce the readability of the site, especially for people with dyslexia.
> * The negative impact of webfonts (additional requests and increased page weight) on page load times may cause a page to go from "very slow" to "unusable", especially in markets of strategic importance for Springer Nature like China. This will in turn impact usage numbers.

performance/font-loading.md
```diff
@@ -0,0 +1,56 @@
...
+  - What impact will the implementation have on the page performance of your site? 
+  
+## Quick Tips
+Here are some quick suggestions for techniques that may prove useful:
+
+### Understand the different user experiences in font loading: FOIT, FOUT and FOIT
+
+- FOIT means flash of invisible text. When web fonts are loading, we hide text so users don’t see anything. We only show the text when web fonts are loaded.
+- FOUT means flash of unstyled text. When web fonts are loading, we show users a system font. When the web font gets loaded, we change the text back to the desired web font.
+- FOFT means flash of faux text. Only load the roman character set initially and load the others later for less page render blocking. "Faux" refers to the fact the browser fakes italics and bold before those character sets have loaded.
```
josebolos, line 25: Suggestion: make definitions a bit clearer by putting "flash of [...] text" in italics, e.g.:

> FOIT means _flash of invisible text_.

performance/font-loading.md
```diff
@@ -0,0 +1,56 @@
...
+  - This is not meant to force you to choose a non-js implementation, but rather for you to ensure you consider what will happen for users in varying circumstances from the beginning.
+- [Our Performance Goals](https://github.com/springernature/frontend-playbook/blob/main/performance/performance-checklist.md)
+  - What impact will the implementation have on the page performance of your site? 
+  
+## Quick Tips
+Here are some quick suggestions for techniques that may prove useful:
+
+### Understand the different user experiences in font loading: FOIT, FOUT and FOIT
+
+- FOIT means flash of invisible text. When web fonts are loading, we hide text so users don’t see anything. We only show the text when web fonts are loaded.
```
josebolos, line 23: > When web fonts are loading, we hide text so users don’t see anything.

Maybe stick a "while" in there, in case people less familiar with technical details end up on this page, something like:

> Text on the page is hidden while webfonts are loading, so users will (or may??) see a page with images but no text. We only show the text when [...]

performance/font-loading.md
```diff
@@ -0,0 +1,56 @@
...
+
+We believe that FOIT should be avoided as it is the worst experience for the user.
+
+### Understand the different implementations available
+
+We highly recommend reading [Zach Leat's comprehensive guide](https://www.zachleat.com/web/comprehensive-webfonts/) to get a good understand of the different techniques available. There is also a good [follow-up article](https://css-tricks.com/the-best-font-loading-strategies-and-how-to-execute-them/) that summarises Zach's findings.
+
+### Serve WOFF and WOFF2 font formats
+
+WOFF2 has the superior compression. However, unfortunately it does not cover all of our Enhanced browsers at the time of writing. Therefore consider serving both formats.
```
josebolos, line 35: Add a link to caniuse, for example:

> [...] unfortunately [it does not cover all of our Enhanced browsers](https://caniuse.com/woff2) at the time of writing [...]

josebolos (commented):
Looks good! I have a few suggestions to make.

performance/font-loading.md
benjclark: Thanks, great suggestion. I have reworded it a bit more as well to try and account for the fact that a person or team might have their own specific reasons for why the webfont might benefit their project, and also to communicate that it is advised to weigh up the pros and cons against each to ensure on balance the pros win.

josebolos (dismissed):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

benjclark:
Will keep this PR open for a total of 2 weeks to give anyone on leave a chance to read and comment.

performance/font-loading.md
```diff
@@ -0,0 +1,60 @@
+# Font Loading
+
+## When to use webfonts
+Before using webfonts it's worth understanding what benefits you believe a webfont will provide your product and assess those against the potential impact they may have on things like accessibility or page speed. For example:
+
+- System fonts are battle tested and were produced with legibility in mind. Using webfonts may reduce the readability of the site, especially for people with dyslexia.
```
dotcode, line 6:
```suggestion
- System fonts are extensively tested and produced with legibility in mind. Using webfonts may reduce the readability of the site, especially for people with dyslexia.
```
Want to avoid militaristic language

performance/font-loading.md
```diff
@@ -0,0 +1,60 @@
...
+- Webfonts can negatively impact page performance (additional http requests and increased page weight) especially in markets of strategic importance for Springer Nature like China. This will in turn impact usage numbers.
+
+Make sure that you feel that on balance the benefit outweighs the impact.
```
dotcode, line 9:
```suggestion
Make sure you can demonstrate that the benefit outweighs the cost.
```
Want to be more prescriptive here

dotcode (changes requested):
Nice work!

performance/font-loading.md
```diff
@@ -0,0 +1,59 @@
+# Font Loading
+
+## When to use webfonts
+Before using webfonts it's worth understanding what benefits you believe a webfont will provide your product and assess those against the potential impact they may have on things like accessibility or page speed. For example:
```
hollsk, line 4:
```suggestion
Before using webfonts it's worth defining what benefits you believe a webfont will provide your product, and assessing those against the potential impact they may have on things like accessibility or page speed. For example:
```

performance/font-loading.md
```diff
@@ -0,0 +1,59 @@
...
+
+- System fonts are extensively tested and produced with legibility in mind. Using webfonts may reduce the readability of the site, especially for people with dyslexia.
```
hollsk, line 6:
```suggestion
- System fonts are extensively tested and produced with legibility in mind. Using webfonts may reduce the readability of the site, for example for people with dyslexia, or low vision.
```

performance/font-loading.md
```diff
@@ -0,0 +1,59 @@
...
+
+WOFF2 has the superior compression. However, unfortunately [it does not cover all of our Enhanced browsers](https://caniuse.com/woff2) at the time of writing. Therefore consider serving both formats. 
+
+### [font-display CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display)
+
+[`font-display`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) is useful and has the benefit of being a technique that does not require javascript and is easy and straight forward to implement.
+
+`font-display: swap` can be used to achieve FOUT which ensures a user can read text with a fallback font whilst a webfont loads.
+
+### Loading fonts with javascript
```
hollsk, line 47:
```suggestion
### Loading fonts with JavaScript
```

performance/font-loading.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Preloading webfonts
+
+If you have to load a font the conventional way using a `<link>` then consider using the [preload technique](https://web.dev/preload-critical-assets/) to ensure the font is requested as early on in the page load process as possible. 
+
+`<link rel="preload" href="HardingText-Regular-Web.woff2" as="font" type="font/woff2" crossorigin>`
+
+### Self host webfonts if possible
+
+Some webfonts prevent you from self hosting because of their licensing so check that first. But if you can, consider self hosting a font. As this will potentially reduce the possible points of failure or poor page performance. And it will open up more options to you in terms of possible font loading implementations.
```
hollsk, line 59:
```suggestion
Some webfonts prevent you from self hosting because of their licensing, so check that first. But if you can, consider self hosting a font. This will potentially reduce the possible points of failure or poor page performance, and it will open up more options to you in terms of possible font loading implementations.
```

hollsk (commented):
tres nice :chefkiss:

made some small suggestions, accept or reject as you prefer :-)

performance/font-loading.md
benjclark: Have reworked this section now so no need to address this suggestion. Have committed your other suggestions though. Thanks!

benjclark:
Excellent discussion on our recommendations around webfonts happened relating to this PR, which can be read here: https://springernature.slack.com/archives/C0GDUEGMS/p1618571720137700

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
+# Font Loading
+
+## When to use webfonts
+
+Webfonts allow us to bring emotion, personality and tone to our sites - which can make them more successful.
+
+However, the task of choosing a webfont and implementing it on your site should be approached cautiously. Webfonts have the potential to impact aspects of your sites such as accessibility and page speed. For example:
+
+- Using webfonts may reduce the readability of the site, for example for people with dyslexia, or low vision.
```
hollsk, line 9:
```suggestion
- Certain webfonts may reduce the readability of the site, for example for people with dyslexia, or low vision.
```

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
...
+- Webfonts can negatively impact page performance (additional http requests and increased page weight) especially in markets of strategic importance for Springer Nature like China. This will in turn impact usage numbers.
+
+Our recommendation is that you use a [system font](https://css-tricks.com/snippets/css/system-font-stack/) instead of a webfont. System fonts are extensively tested, incur no performance penalty, and are produced with legibility in mind.
```
hollsk, line 12:
```suggestion
Our recommendation is that you prefer a [system font](https://css-tricks.com/snippets/css/system-font-stack/) over a webfont. System fonts are extensively tested, incur no performance penalty, and are produced with legibility in mind.
```

Though maybe "prioritise" is a better word than "prefer"?

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
...
+
+If a webfont is needed then close collaboration between User Experience and Accessibility specialists should be sought to ensure the webfont chosen is satisfactory from both typographic and accessibility perspectives.
```
hollsk, line 14:
```suggestion
If you need a webfont, then close collaboration between User Experience and Accessibility specialists should be sought to ensure your chosen webfont is satisfactory from both typographic and accessibility perspectives.
```

hollsk (commented):
gosh yes this is nice

performance/font-loading.md
benjclark: Yeah I like prioritise better than prefer in that sentence 👍

dotcode: `**We recommend to use a [system font](https://css-tricks.com/snippets/css/system-font-stack/) instead of a webfont**. System fonts are extensively tested, and incur no performance penalty.`

Highlighting the recommendation.

Also taking into account @foxintherain's feedback about Garamond and legibility, we can't say that "system fonts are legible", per se.

performance/font-loading.md
dotcode: `If a webfont is needed then close collaboration between user experience, accessibility, and web performance specialists must be sought to ensure the choice is satisfactory from typographic, accessibility, and performance perspectives.

As a Frontend Developer, it's your role to provide technical guidance and feedback, not to choose the font.`

Changed the "should seek" to a "must seek", included web performance, and added a bit of guidance around the role a FED plays in this.

dotcode (commented):
Thanks Ben … just a couple of tweaks. What do you reckon?

performance/font-loading.md
moddular: Garamond isn't a system font though

dotcode: This might be a pertinent point, knew I should've stuck with Webdings 😂

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Serve WOFF and WOFF2 font formats
+
+WOFF2 has the superior compression. However, unfortunately [it does not cover all of our Enhanced browsers](https://caniuse.com/woff2) at the time of writing. Therefore consider serving both formats. 
+
+### [font-display CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display)
+
+[`font-display`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display) is useful and has the benefit of being a technique that does not require JavaScript and is easy and straight forward to implement.
+
+`font-display: swap` can be used to achieve FOUT which ensures a user can read text with a fallback font whilst a webfont loads.
```
josebolos, line 50: I know this paragraph is not saying that `font-display: swap` is the best or the only useful value for `font-display`, but given that it's the only one mentioned, I feel like we may want to clarify that all possible `font-display` values have serious downsides. Although we're risking this PR becoming comically large if we include that.

For example, `swap` (and `block` too, I think) will cause the whole page to be rendered completely twice, doubling the amount of paint events and increasing. The geometry is probably not the same between fonts, so the shifts in the layout may be annoying or jarring for some people.

Gov.uk uses `fallback`, and, there's an excellent article from Matt Hobbs showing the whole decision process of why they decided to use that value:
https://calendar.perfplanet.com/2020/a-font-display-setting-for-slow-connections/

I'm not saying that `fallback` is best by the way, just that this is a bit too complex for a short mention so maybe we can just let others do all the hard work and link to that post, or a similar one?

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
...
+  - This is not meant to force you to choose a non-js implementation, but rather for you to ensure you consider what will happen for users in varying circumstances from the beginning.
+- [Our Performance Goals](https://github.com/springernature/frontend-playbook/blob/main/performance/performance-checklist.md)
+  - What impact will the implementation have on the page performance of your site? 
+  
+## Quick Tips
+Here are some quick suggestions for techniques that may prove useful:
+
+### Understand the different user experiences in font loading: FOIT, FOUT and FOIT
+
+- FOIT means _flash of invisible text_. Text on the page is hidden while webfonts are loading, so users will see a page with images but no text. We only show the text when web fonts are loaded.
```
sonniesedge, line 32: :twitching_eye_emoji: web font -> webfont

performance/font-loading.md
```diff
@@ -0,0 +1,64 @@
...
+- FOUT means _flash of unstyled text_. When web fonts are loading, we show users a system font. When the web font gets loaded, we change the text back to the desired web font.
```
sonniesedge, line 33: :twitching_eye_emoji: web font -> webfont

sonniesedge:
This is 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 .

performance/font-loading.md
benjclark: It's probably true though that not all system fonts could be said to have good legibility. So safest to take that bit out I think.

foxintherain: My jokes aside about Garamond - I think the main system fonts for OS' (Windows, Apple, Google) absolutely have been tested and very legible - so I think you can leave it in! Maybe I'm getting confused here between browser safe and system os

benjclark: I dunno, I can't verify that all system fonts meet that sweeping statement of being "produced with legibility in mind" so it does feel safer just to remove. Think it's better without.

performance/font-loading.md
```diff
@@ -0,0 +1,67 @@
...
+  - Does the implementation work in conjunction with the Cut The Mustard technique?
+- [Progressive Enhancement](https://github.com/springernature/frontend-playbook/blob/main/practices/progressive-enhancement.md)
+  - This is not meant to force you to choose a non-js implementation, but rather for you to ensure you consider what will happen for users in varying circumstances from the beginning.
+- [Our Performance Goals](https://github.com/springernature/frontend-playbook/blob/main/performance/performance-checklist.md)
+  - What impact will the implementation have on the page performance of your site? 
+  
+## Quick Tips
+Here are some quick suggestions for techniques that may prove useful:
+
+### Understand the different user experiences in font loading: FOIT, FOUT and FOIT
```
axemonkey, line 33: Typo

axemonkey, line 33: Typo - second **FOIT** should be **FOFT**

benjclark: Good spot. 😂 
It's amazing how many times you can read something and not spot an obvious mistake like that.

hollsk (approved):
🎷

alexkilgour (approved):
Nice Work!

dotcode (approved):
Nice work @benjclark and all!

PR #380, opened 2021-04-21
Update the link to the axe browser extension

hollsk:
The main page on Deque's website that describes what axe is has a big button to install the extension on Chrome. The extension is in fact available for multiple browsers, it's just that this information (and the relevant links to each extension store) is buried away where nobody will ever find it. 

This PR makes it explicit that the extension is available for multiple browsers, and where to find the right one.


benjclark (approved):
👍

PR #381, opened 2021-05-05
recommend axe Accessibility Linter for VS Code

hollsk:
Deque have released a linter for VS Code, so now you can catch problems more earlier in yer dev process. This PR adds a recommendation that you do so.

Here is the announcement: https://www.deque.com/blog/shift-further-left-with-deques-axe-linter-for-vs-code/


josebolos (approved):
:+1:

PR #382, opened 2021-05-31
Some words about for...of loops for client side JS

morgaan:


tulipdexter:
I wonder if, instead of calling out `for...of` loops as their own section, it's worth having a section like "Be careful when transpiling" or something similar. The reason to not use `for...of` loops if you use IE is because the transpiled size is huge 🙈  and I think that applies to more than just `for...of`.
Screenshot 2021-06-01 at 09 05 51

morgaan:
@tulipdexter This makes perfect sense. I honestly had no idea where to put it, so I went for the Loops section. Let me try to find a spot for that section and stick this into there rather. They may be other stuffs we may stick in there.

morgaan:
@tulipdexter I gave it a try. Would you mind have another look? Feel free to fix my non native speaker english :p

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
   - It can expose functions and/or prototypal classes
 
 An example utility might be a function to make a string title-case.
+
+### Be careful when transpiling
+
+As exciting it can be to use modern Javascript syntax, it remains our
```
jpw, line 877:
```suggestion
As exciting as it can be to use modern Javascript syntax, it remains our
```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+responsibility as web developers to make sure our benefits do not harm our user
+experience. Any negative impact on performance affects the user experience.  
+Transpiling does increase Javascript bundles size.
```
jpw, line 880:
```suggestion
Transpiling does increase the JavaScript bundle size.  
```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+We do this:
+
+```js
+const list = document.querySelectorAll('input[type=checkbox]');
+Array.prototype.forEach.call(list, function (checkbox) {
+    // ...
+});
+```
+
+If you would run the above "We don't do this" snippet into [Babel REPL](https://babeljs.io/repl/),
```
jpw, line 905:
```suggestion
If you would run the above "We don't do this" snippet in [Babel REPL](https://babeljs.io/repl/),
```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+using our [browserslist](https://github.com/springernature/frontend-playbook/blob/main/practices/graded-browser-support.md#browerslist),
+it transpiles down to 23 lines where in opposition the "We do this" snippet
```
jpw, line 907:
```suggestion
it transpiles down to 23 lines, whereas the "We do this" snippet
```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+transpiles down to 5 lines.  If you would multiply this by the amount of
```
jpw, line 908:
```suggestion
transpiles down to 5 lines.  If you multiply this by the amount of
```

jpw (changes requested):
minor grammatical etc tweaks

javascript/house-style.md
sonniesedge, line 880:
```suggestion
As enticing it can be to use modern Javascript syntax, it remains the developers responsibility to ensure the developer experience does not negatively impact the user experience, e.g. by decreasing performance. 

Transpiling can easily increase the JavaScript bundle size, by incorporating polyfills to for supported browsers.

```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+
+### Be careful when transpiling
+
+As exciting it can be to use modern Javascript syntax, it remains our
+responsibility as web developers to make sure our benefits do not harm our user
+experience. Any negative impact on performance affects the user experience.  
+Transpiling does increase Javascript bundles size.  
+We therefore advise to monitor, with tools like [Babel
+REPL](https://babeljs.io/repl/), the footprint of your code and reconsider your
+options, most especially if the only purpose is syntax sugar.
```
sonniesedge, line 883:
```suggestion
There are two ways of avoiding an increase in bundle size:

1. Monitoring, with tools like [BabelREPL](https://babeljs.io/repl/), which show the size of your transpiled code.
1. Using "older" JavaScript syntax to achieve the same effect. An example of this is given below.
```

javascript/house-style.md
```diff
@@ -871,3 +871,41 @@ The `utils` directory is used to house JavaScript written for the project that d
...
+});
+```
+
+If you would run the above "We don't do this" snippet into [Babel REPL](https://babeljs.io/repl/),
+using our [browserslist](https://github.com/springernature/frontend-playbook/blob/main/practices/graded-browser-support.md#browerslist),
+it transpiles down to 23 lines where in opposition the "We do this" snippet
+transpiles down to 5 lines.  If you would multiply this by the amount of
+occurences of `for...of` in your code base, this can quickly get out of control.
+
+You can find more in depth details about this use case in [Transpiled for-of Loops are Bad for the Client, by Dave Ruppert](https://daverupert.com/2017/10/for-of-loops-are-bad/).
```
sonniesedge, line 911: In depth details: 
- [Transpiled for-of Loops are Bad for the Client, by Dave Ruppert](https://daverupert.com/2017/10/for-of-loops-are-bad/).
- [Avoiding Babel’s Production Bloat](https://webreflection.medium.com/avoiding-babels-production-bloat-d53eea2e1cbf)

sonniesedge:
I have put a lot of annoying suggestions in there that make it look like I don't like what @morgaan wrote WHEN IN FACT THIS IS NOT TRUE I'M JUST AN ANNOYING PEDANT.

morgaan:
I incorporated your suggestions dear @sonniesedge and @jpw. Thank you to you two for that!

jpw (approved):

PR #389, opened 2021-09-29
Adds section for Sass variable naming convention in how-we-write-css.md

benjclark:
As per discussion in PR https://github.com/springernature/frontend-toolkits/pull/577#issuecomment-930092088

It would be good to define our Sass variable naming convention in the Playbook

### DIFF
--- a/css/how-we-write-css.md
+++ b/css/how-we-write-css.md
@@ -55,7 +55,7 @@ It can be useful to use numbers at the start of your folder names so that the or
 40-base
 50-components
 60-utilities
-```
+``` 
 
 ### Wider architecture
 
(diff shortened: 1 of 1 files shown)


css/how-we-write-css.md
```diff
@@ -107,6 +107,28 @@ We follow the following rules when writing class names:
 * All words are separated by a dash/hyphen
 * BEM naming conventions are followed (read more about our [approach to writing BEM](bem-css.md))
 
+## Sass variable naming
+
+We follow the following rules when writing Sass variable names:
```
jon-stevens, line 112:
```suggestion
We follow these rules when writing Sass variable names:
```

sturobson:
When values exist (hello default context / brand context Sass maps) … should this add something along the lines of

"if available use existing context Sass variables instead of hard coded values"

jon-stevens (approved):

PR #394, opened 2021-11-10
Move ARC toolkit into the right section

hollsk:
For some mysterious reason I put the ARC toolkit into the manual testing tools section, instead of in the automated tools section where it belongs 🤷‍♀️


besing (approved):
👏  Well done @hollsk

PR #398, opened 2022-01-06
Adds sub-section for file naming in writing css guidelines

benjclark:
This should help to create consistency in how we name scss files in our scss architecture.

The guidance is general and tries to apply it to js files as well. However, noted that this is going in to the 'how-we-write-css' guidance - which isn't a good place to describe this.

Any thoughts on where the same js file naming guidance could go instead?

### DIFF
--- a/css/how-we-write-css.md
+++ b/css/how-we-write-css.md
@@ -57,6 +57,55 @@ It can be useful to use numbers at the start of your folder names so that the or
 60-utilities
 ``` 
 
+### File naming
+
+Name your files as the component name with the namespace subtracted.
+
+Keep your scss or js in one file if it is all mandatory. This reduces the number of imports consumers will need to make in their applications and prevents any confusion where readers may think something is optional.
+
+We _don't_ do this:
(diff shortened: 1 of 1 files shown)


css/how-we-write-css.md
```diff
@@ -57,6 +57,55 @@ It can be useful to use numbers at the start of your folder names so that the or
...
+
+We _don't_ do this:
+```
+global-popup
+     __tests__
+         global-popup.spec.js
+     js
+         global-popup.js
+     scss
+         10 settings
```
ricky-lau, line 74: the folders `10 settings` and `50 components` seemed to be missing the hyphen

ricky-lau (approved):

benjclark:
I have realised this is better in the frontend-elements-docs repo where this is already guidance for this. See https://github.com/springernature/frontend-elements-docs/pull/13

sturobson:
Holistically, should the front-end playbook not be part of the EDS?

PR #399, opened 2022-01-13
Embedding MathML and SVG in HTML

jpw:
- adds "Closing elements in embedded documents" section
 - other minor wording and whitespace tweaks

### DIFF
--- a/markup/house-style.md
+++ b/markup/house-style.md
@@ -1,14 +1,16 @@
 # Markup
 
-This document outlines the way we write markup and why. See our [house style document](../practices/house-style.md) to understand the rationale behind enforcement of style. 
+This document outlines the way we write markup and why. See our [house style document](../practices/house-style.md) to understand the rationale behind enforcement of style.
 
 - [HTML vs XHTML](#html-vs-xhtml)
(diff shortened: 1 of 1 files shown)


moddular:
In most of our article/book/chapter content we tend to use TeX rather than MathML for equations, so I'm not entirely sure about it. Though I would suspect given that much of this type of content is generated by an XSL transform on an XML document we'd end up with namespace gubbins and `/>` being output.

The SVG stuff sounds right, though again I'd probably caveat that for places where that comes from some XSLT.

jpw:
Thanks @moddular. I also suspect many instances of this come from XSLT tools. IIRC XSLT tools can be configured to output `html` serialisation but in practice, people tend not to 😞

I suppose SVG tools default to XML serialisation so as to be safe in strict environments (SVG docs) and tend not to worry about embedding in HTML as it "will probably work in browsers because they can mainly handle both serialisations".

jpw:
How did you get on with this in the end @moddular  ?

moddular (approved):
All sounds good to me. The XSLT has been updated to output an HTML serialisation.

PR #402, opened 2022-01-26
add a page about alt text

hollsk:
We get a lot of people asking us about alt text. This is a first attempt at pulling together info from different sources into a single document we can hand over to people who ask us for guidance on alt text. 

About 80% of the document was borrowed from the Nature magazine editorial staff guide. I'm not pretending this is a complete document - for example, I haven't included anything about infographics at this stage. Can add other sections as we need them.


accessibility/alt-text.md
```diff
@@ -0,0 +1,162 @@
...
+    - [Good alt text](#good-alt-text-2)
+    - [Bad alt text](#bad-alt-text-2)
+  - [Headshots](#headshots)
+    - [Bad alt text](#bad-alt-text-3)
+    - [Good alt text](#good-alt-text-3)
+
+
+## What is alt text? 
+
+It's text used to describe the appearance and function of an image on a page, as a textual alternative to non-text content. You don't normally see it when you're browsing page as a sighted web user.
```
jon-stevens, line 26:
```suggestion
It's text used to describe the appearance and function of an image on a page, a textual alternative to non-text content. You don't normally see it when you're browsing a page as a sighted web user. 
```

PR #403, opened 2022-02-01
Adds info re self-hosting fonts

jon-stevens:
blep-font

### DIFF
--- a/performance/font-loading.md
+++ b/performance/font-loading.md
@@ -62,6 +62,6 @@ If you have to load a font the conventional way using a `<link>` then consider u
 
 `<link rel="preload" href="HardingText-Regular-Web.woff2" as="font" type="font/woff2" crossorigin>`
 
-### Self host webfonts if possible
+### Self host webfonts
 
-Some webfonts prevent you from self hosting because of their licensing so check that first. But if you can, consider self hosting a font. This will potentially reduce the possible points of failure or poor page performance, and it will open up more options to you in terms of possible font loading implementations.
+Any webfonts you use must be self-hosted - avoid using a font-hosting service. Relying on external services for fonts adds another possible point of failure, risks poor page performance, and importantly, it risks passing personally identifiable information from your users to those font-hosting services.
--- a/performance/font-loading.md
+++ b/performance/font-loading.md
@@ -68,6 +68,6 @@ Any webfonts you use must be self-hosted - avoid using a font-hosting service.
 
 Every major browser now implements HTTP cache partitioning in order to prevent leaking of the users' browser history, and avoid cross-site tracking and cross-site search attacks. This means that there's no performance benefit from using an external service to host our webfonts.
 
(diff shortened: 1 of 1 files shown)


axemonkey (approved):
LGTM

dotcode:
Do not merge, yet

performance/font-loading.md
```diff
@@ -62,6 +62,6 @@ If you have to load a font the conventional way using a `<link>` then consider u
 
 `<link rel="preload" href="HardingText-Regular-Web.woff2" as="font" type="font/woff2" crossorigin>`
 
-### Self host webfonts if possible
+### Self host webfonts
 
-Some webfonts prevent you from self hosting because of their licensing so check that first. But if you can, consider self hosting a font. This will potentially reduce the possible points of failure or poor page performance, and it will open up more options to you in terms of possible font loading implementations.
+Any webfonts you use must be self-hosted - avoid using a font-hosting service. Relying on external services for fonts adds another possible point of failure, risks poor page performance, and importantly, it risks passing personally identifiable information from your users to those font-hosting services.
```
josebolos, line 67:
```suggestion
Any webfonts you use must be self-hosted - avoid using a font-hosting service.

Every major browser now implements HTTP cache partitioning in order to prevent leaking of the users' browser history, and avoid cross-site tracking and cross-site search attacks. This means that there's no performance benefit from using an external service to host our webfonts.

Using a external font-hosting service can actually worsen performance in important markets like China, as opening a connection to a new host can sometimes be very slow, or even fail completely, due to the filtering done by the Great Firewall.

But most importantly, using external font-hosting services we risk passing personally identifiable information (PII) from your users to those font-hosting services, increasing the risk of litigation.
```

josebolos: Added a bit more of context based on @dotcode's suggestions. I'm unsure about the "increasing the risk of litigation" at the end - @dotcode Do you think that's appropriate for this document?

dotcode: Maybe, but not quite yet. Currently talking to Legal about that German Google Fonts case.

SamLeach: Hi. Just curious about not hosting static assets on edge CDNs. Apart from specific cases like China, do you have any links I can read? I was under the impression that edge CDNs were more performant.

jpw: Hi Sam!

You are right in saying that using CDNs is indeed more performant. 

In this context "self-hosting" means using our origin servers, cached on the edge by a CDN. I appreciate this can be a bit confusing, we should probably clarify that.

However, using for example Google as an origin server, _and_ CDN (as per typical Google font usage) may well have different legal implications.

Hosting resources in China introduces yet more legal complexities which are probably nest not discussed in an open source repo, as I'm sure you understand 😉

jpw:
@dotcode @josebolos are there issues preventing this being merged?

josebolos:
I'm happy with it, but I think that @dotcode had some concerns.

jpw:
LGTM!

jpw (approved):
Thank you! 🐱

PR #404, opened 2022-02-11
Improve images doc and add new image formats

josebolos:
* Add a table of contents
* Improve the text in the Introduction
* Make "Picking the right image format" a separate section
* Add MP4 to the same section
* Add modern formats like WEBP, AVIF and JPEG-XL
* Explicitly reject the use of GIFs ever, for any reason, to anyone for any reason ever, no matter what, no matter where, or who, or who you are with, or where you are going, or where you've been... ever, for any reason whatsoever...
* Make Optimisation section clearer
* Remove named image servers and services to avoid people thinking that we're endorsing them


performance/images.md
```diff
@@ -28,48 +49,41 @@ When encoding JPEG images:
...
-* Strip the metadata from the image. This includes:
-  * EXIF data
-  * Copyright information (unless required)
-  * GPS data
-  * Colour profiles
-  * Thumbnails in JPEGs
-  * Comments
-  * Etc.
-* Set appropriate cache headers for the images. (See [Serving images](#serving-images), below.)
+Images can frequently make up a significant amount of a page's size, so it's important to optimise them before they are delivered to the users in order to ensure that our sites stay performant and we're not preventing users from accessing the site due to the increased page weight. Common optimisations include:
```
jon-stevens, line 60:
```suggestion
Images can contribute significantly to a page's size, so it's important to optimise them before they are delivered to the users. This helps ensure that our sites stay performant and we're not preventing anyone from accessing them due to increased page weight. Common optimisations include:
```

performance/images.md
```diff
@@ -1,22 +1,43 @@
...
+* Use [JPEG](https://en.wikipedia.org/wiki/JPEG) for still images, photographs, and in any other cases where using a PNG would result in an image too large to be delivered over the network effectively.
+* Use [PNG](https://en.wikipedia.org/wiki/Portable_Network_Graphics) for illustrations, graphs, flat-colour graphics, images containing transparency (which JPEG doesn't support), and images containing text.
+* Use [MP4](https://en.wikipedia.org/wiki/MPEG-4_Part_14) for any kind of animated images.
+* Do not use [GIFs](https://en.wikipedia.org/wiki/GIF). For animated images, the file size of GIF files, usually an order of magnitude bigger than MP4s of similar quality, may be a barrier preventing users from accessing our content, specially in markets where slower connections and low performing devices are more common. For static images with transparency, an 8-bit PNG can be used to deliver an indexed colour image, with binary transparency, [at a smaller size than a GIF](https://helpx.adobe.com/photoshop-elements/using/optimizing-images-gif-or-png.html).
 
-## Formats
+A number of more modern image formats can be also used to deliver images:
```
jon-stevens, line 33:
```suggestion
A number of more modern image formats can also be used to deliver images:
```

jon-stevens (approved):
🥳

jon-stevens (approved):

performance/images.md
```diff
@@ -1,22 +1,43 @@
...
+* Use [SVG](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics) for icons, logos, and other images with sharply defined edges.
+* Use [JPEG](https://en.wikipedia.org/wiki/JPEG) for still images, photographs, and in any other cases where using a PNG would result in an image too large to be delivered over the network effectively.
+* Use [PNG](https://en.wikipedia.org/wiki/Portable_Network_Graphics) for illustrations, graphs, flat-colour graphics, images containing transparency (which JPEG doesn't support), and images containing text.
+* Use [MP4](https://en.wikipedia.org/wiki/MPEG-4_Part_14) for any kind of animated images.
+* Do not use [GIFs](https://en.wikipedia.org/wiki/GIF). For animated images, the file size of GIF files, usually an order of magnitude bigger than MP4s of similar quality, may be a barrier preventing users from accessing our content, specially in markets where slower connections and low performing devices are more common. For static images with transparency, an 8-bit PNG can be used to deliver an indexed colour image, with binary transparency, [at a smaller size than a GIF](https://helpx.adobe.com/photoshop-elements/using/optimizing-images-gif-or-png.html).
```
dotcode, line 31: What about images without transparency … PNGs still smaller?

hollsk, line 31:
```suggestion
* Do not use [GIFs](https://en.wikipedia.org/wiki/GIF). The file size of animated GIF images is usually an order of magnitude bigger than MP4s of similar quality. They can be a barrier that prevents users from accessing our content, especially in markets where slower connections and low performing devices are more common. For static images with transparency, an 8-bit PNG can be used to deliver an indexed colour image, with binary transparency, [at a smaller size than a GIF](https://helpx.adobe.com/photoshop-elements/using/optimizing-images-gif-or-png.html).
```

Turns one long long ~~man~~ sentence into two shorter ones

performance/images.md
```diff
@@ -28,48 +49,41 @@ When encoding JPEG images:
...
 
 ## Serving images
 
-We work under the assumption that images are encoded only once but decoded many times. Therefore, encoding times are usually secondary to the savings in e.g. file size.
+We work under the assumption that images are encoded only once but decoded many times. Therefore, increased encoding times, while undesirable, are usually secondary to other concerns like making the image easier to decompress, or having a smaller file size.
 
 When optimising non-content images that are going to be committed to a repo, make sure you always use the maximum settings if you're using lossless optimisations.
 
-For content images, we encourage using an image server like [Immagine](https://github.com/springernature/immagine) or an image service like [Kraken.io](https://kraken.io/) in order to:
+For content images, we encourage using an on-premise or cloud-based image server in order to:
```
hollsk, line 80: should we mention Spindoctor here? or is it too platform-specific?

performance/images.md
```diff
@@ -28,48 +49,41 @@ When encoding JPEG images:
...
+* Ensure that all images are optimised automatically.
+* Ensure that the appropriate cache headers are always set.
 
-Also, future improvements to these tools will automatically be reflected in all of the images being served, instead of them having to be re-encoded.
+Additionally, future improvements to these tools will result in improvements to the images being served automatically, without the images having to be manually re-encoded.
 
 ## Comparing the quality of images
 
-Besides visual inspection, if you need to compare the quality of two different image compression formats, algorithms or quality settings, you may want to use something like the [SSIM index](http://en.wikipedia.org/wiki/Structural_similarity) to get an objective idea of the trade-off between quality and byte size.
+When comparing the output of two different image compression formats, algorithms, or quality settings, [tools](https://github.com/kornelski/dssim) that calculate the similarity between images with algorithms like [SSIM](http://en.wikipedia.org/wiki/Structural_similarity) and DSSIM can be used to get an objective idea of the trade-off between quality and byte size.
```
hollsk, line 89:
```suggestion
When comparing the output of two different image compression formats, algorithms, or quality settings, [tools that calculate the similarity between images](https://github.com/kornelski/dssim) with algorithms like [SSIM](http://en.wikipedia.org/wiki/Structural_similarity) and DSSIM can be used to get an objective idea of the trade-off between quality and byte size.
```

performance/images.md
ricky-lau: I wanted to mention the quality settings from Spindoctor 😬 but this could also be something to reference about the quality setting.

From the Spindoctor README:
>By default, PNGs will be output at compression quality of 95, and JPEGs at 90 (for lossless sources such as PNG or TIFF) or 80 (for lossy sources).

Those seem quite high? This is just a feeling and I don't know how those original settings came about. I suppose with photographic images you could get away with lower but it may be more noticeable against figure images with graphs.

performance/images.md
josebolos: It should, yes. The reason why the text only mentions the transparency use case is because GIFs have been traditionally used for this purpose. But from the Adobe link mentioned:

> The PNG‑8 format uses more advanced compression schemes than GIF does, and a PNG‑8 file can be 10% to 30% smaller than a GIF file of the same image, depending on the image’s color patterns.

This is independent from the image using transparency or not.

performance/images.md
```diff
@@ -1,22 +1,43 @@
...
+* [AVIF](https://en.wikipedia.org/wiki/AVIF) is suitable for static and animated images, and supports both lossy and lossless compression with better compression efficiency and better detail preservation than other formats. It's [supported by Chrome and Firefox](https://caniuse.com/avif), and can be used by providing another better supported format as a fallback. Compression (encoding) of images can be significantly slower than with other formats.
+* [JPEG-XL](https://en.wikipedia.org/wiki/JPEG_XL) is in the process of becoming a standard and supports both lossy and lossless compression. It's [more efficient than other formats while being usually faster](https://cloudinary.com/blog/how_jpeg_xl_compares_to_other_image_codecs
+) both when compressing and decompressing than JPEG. [No browser officially supports it](https://caniuse.com/jpegxl) at the time of writing.
```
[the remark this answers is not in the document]
josebolos: 👍

performance/images.md
josebolos: Spindoctor is not public or open source, so I thought it might be confusing for external readers to see a product mentioned that is not publicly available. I also don't feel like we should endorse any products, which is why I remove the reference to specific services.

> Those seem quite high? This is just a feeling and I don't know how those original settings came about. I suppose with photographic images you could get away with lower but it may be more noticeable against figure images with graphs.

Correct. Defaults were adjusted in April 2020. Happy to discuss specifics in Slack.

performance/images.md
```diff
@@ -1,22 +1,43 @@
...
+* Use [MP4](https://en.wikipedia.org/wiki/MPEG-4_Part_14) for any kind of animated images.
+* Do not use [GIFs](https://en.wikipedia.org/wiki/GIF). The file size of animated GIF images is usually an order of magnitude bigger than MP4s of similar quality. They can be a barrier that prevents users from accessing our content, especially in markets where slower connections and low performing devices are more common. For static images with transparency, an 8-bit PNG can be used to deliver an indexed colour image, with binary transparency, [at a smaller size than a GIF](https://helpx.adobe.com/photoshop-elements/using/optimizing-images-gif-or-png.html).
 
-## Formats
+A number of more modern image formats can also be used to deliver images:
 
-### Which format to use
+* [WEBP](https://en.wikipedia.org/wiki/WebP) is a replacement for JPEG, PNG, and GIFs, but improvements to encoders like mozjpeg means that WEBP file sizes are only marginally smaller than JPEG. It's supported by almost all major browsers ([Safari supports it partially](https://caniuse.com/webp)) and can be used by providing another better supported format as a fallback.
```
dotcode, line 35:
```suggestion
* [WEBP](https://en.wikipedia.org/wiki/WebP) is a replacement for JPEG, PNG, and GIFs, but improvements to encoders like mozjpeg means that WEBP file sizes are only marginally smaller than JPEG. It's supported by almost all major browsers ([Safari supports it partially](https://caniuse.com/webp)) and can be used only as an enhancement to another better supported format.
```
Using the language of progressive enhancement rather than graceful degradation.

performance/images.md
```diff
@@ -1,22 +1,43 @@
...
+* [AVIF](https://en.wikipedia.org/wiki/AVIF) is suitable for static and animated images, and supports both lossy and lossless compression with better compression efficiency and better detail preservation than other formats. It's [supported by Chrome and Firefox](https://caniuse.com/avif), and can be used by providing another better supported format as a fallback. Compression (encoding) of images can be significantly slower than with other formats.
```
dotcode, line 36:
```suggestion
* [AVIF](https://en.wikipedia.org/wiki/AVIF) is suitable for static and animated images, and supports both lossy and lossless compression with better compression efficiency and better detail preservation than other formats. It's [supported by Chrome and Firefox](https://caniuse.com/avif), and can be used only as an enhancement to another better supported format. Compression (encoding) of images can be significantly slower than with other formats.
```
Using the language of progressive enhancement  rather than graceful degradation.

dotcode (changes requested):

josebolos:
I think that's all of the comments addressed! Please do let me know if not!

dotcode (approved):
Thanks for incorporating those changes Jose :)

PR #406, opened 2022-03-08
Updates graded browser support doc after IE11 reclassification

benjclark:
- Graded browser support page updated to remove IE11 (and IE10) from Enhanced browser list and caveat section
- Cut The Mustard media query updated. This has been taken from [Andy Kirk's guide](https://github.com/Fall-Back/CSS-Mustard-Cut#pcs-cut-prefers-color-scheme) and is considered the most up to date media query. However the matched browser versions are still a long way off the latest versions of each browser.
- Removes mention of "latest stable" and "latest stable -1" as we are unable to implement those (see above re media query) so was potentially causing unnecessary confusion
- Updates the browserlist config

I spent a day playing with media queries to try and find something that matched recent versions of each browser. The closest I got was `only print, screen and (forced-colors:none)` which had the following support:
image
which is obviously no good. One of the difficulties is that we are defining media queries that are to be applied to @media therefore only [media features](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media_features) are applicable for use in the query. Therefore the one used in this PR (`prefers-color-scheme`) is the best we have using this technique at this time. It matches: `Edge 79+, Chrome 76+, Firefox 67+, Opera 62+, Safari 12.1+, iOS 13+, Android 91+`

Ideally going forward we should try and find a better approach to CTM, one that allows us to get back to a policy of "latest stable" and "latest stable -1" for our enhanced browsers. This PR allows us to move forward meantime.

Note - some of the modern CSS features like Logical Properties (`Edge 89+, FF 66+, Chrome 89+, Safari 15+, Opera 76+, Android 98+`) cannot be used in Enhanced styles using this CTM implementation.

### DIFF
--- a/markup/house-style.md
+++ b/markup/house-style.md
@@ -1,14 +1,16 @@
 # Markup
 
-This document outlines the way we write markup and why. See our [house style document](../practices/house-style.md) to understand the rationale behind enforcement of style. 
+This document outlines the way we write markup and why. See our [house style document](../practices/house-style.md) to understand the rationale behind enforcement of style.
 
 - [HTML vs XHTML](#html-vs-xhtml)
-	- [Validation](#validation)
-	- [Do not omit optional end tags](#do-not-omit-optional-end-tags)
-	- [Implicitly close void elements](#implicitly-close-void-elements)
-	- [Form submit controls](#form-submit-controls)
+  - [We author HTML 5](#we-author-html-5)
+  - [Validation](#validation)
+  - [Do not omit optional end tags](#do-not-omit-optional-end-tags)
+  - [Implicitly close void elements](#implicitly-close-void-elements)
+    - [Closing elements in embedded documents](#closing-elements-in-embedded-documents)
+  - [Form submit controls](#form-submit-controls)
 - [Semantics](#semantics)
-	- [Follow the HTML 4 outline model for heading levels](#follow-the-html-4-outline-model-for-heading-levels)
+  - [Follow the HTML 4 outline model for heading levels](#follow-the-html-4-outline-model-for-heading-levels)
 - [Templates](#templates)
 
 ## HTML vs XHTML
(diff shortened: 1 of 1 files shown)


practices/graded-browser-support.md
```diff
@@ -115,10 +99,9 @@ The following `.browserslistrc` file ([a standard way of sharing target browser
 
 ```nanorc
 defaults
```
ricky-lau, line 101: Add `not ie 11` as `defaults` still includes IE11. [List of browsers without `not ie 11`](https://browserslist.dev/?q=ZGVmYXVsdHMsIGZmID4gNjYsIGNocm9tZSA%2BIDc1LCBzYWZhcmkgPiAxMSwgZWRnZSA%2BIDc4LCBvcGVyYSA%2BIDYx)

benjclark: Ah ok , well spotted 👍 

That site is handy. Browserlist's README says defaults means: `> 0.5%, last 2 versions, Firefox ESR, not dead` and I think at the time I thought ie11 was either `dead` or `< 0.5%`

besing (approved):
LGTM

aliuk2012:
I think it might be worth adding some documentation (if there isn't any) about how this might affect ES5 and compiling js to various ES versions.

Should we still be compiling our js to ES5 or for Core experience do we just not serve up any js?

besing:
@aliuk2012 According to the [playbook](https://github.com/springernature/frontend-playbook/blob/main/practices/graded-browser-support.md#our-criteria-for-browser-support) we don't serve JS to core browsers. We use a JS `window.matchMedia` method to apply our CSS-cut-the-mustard to JS as well, i.e. not serving our JS bundle to those same browsers.

aliuk2012:
🤦  Oh of course, we were only compiling to es5 because of IE 11 was considered "Enhanced"

PR #407, opened 2022-03-23
Clarifying legal stuff

dotcode:
1. The EU directive applies to European states, we are not a European state, so removed reference to it.
2. *We* can break the law, our websites cannot, so clarified that. Much as we may desire to send particular websites to jail, those days are yet to come (on the upside, we can currently send our colleagues and friends and families to jail instead, so all is not lost).  
3. Being picky, but we cannot say that we MUST comply with ALL of the listed laws. Obviously we want to go _beyond_ the laws, and in any one particular circumstance it's almost certain that at least one of those laws will apply, but in terms of globally applicable guidance, we're restricted to saying we SHOULD comply.


sonniesedge:
> The EU directive applies to European states, we are not a European state, so removed reference to it.

Who's "we"? 😉

sonniesedge:
oh, you're on about _are we a state_, rather than which state we're in? Sorry MISREAD STAND DOWN ETC

dotcode:
Yeah, correct, I mean … there are plenty of signs that the world is heading towards the future posited in [Snow Crash](https://en.wikipedia.org/wiki/Snow_Crash), but until we have the Springer Nature Franchise-Organized Quasi-National Entity established, we can't really consider ourselves to be in the position of issuing passports, collecting taxes, etc.

hollsk:
I know it's merged and the moment is passed, but I have just one question:

We're not a US state either, so should we also remove the bit about Section 508/Rehabilitation Act?

PR #437, opened 2022-10-04
Link underline pointers

Heydon:
Following the open-space discussion (https://springernature.slack.com/archives/C1HGRAQ8G/p1664457489573739) this fulfills one of the actions agreed to be undertaken by the Elements team: to add specific guidance to the playbook about link underlines RE WCAG 1.4.1.

### DIFF
--- a/accessibility/common-remediations/colours.md
+++ b/accessibility/common-remediations/colours.md
@@ -17,12 +17,14 @@ Examples may include (but are not limited to):
 * Changing the background colour of an input field to red to indicate an error without also providing an error message
 * Changing the colour of an icon or text string to indicate some kind of status
 * Using a graph or a chart where colour is the only way to indicate different aspects of the data
+* Removing underlines from hyperlinks, leaving just colour to differentiate them from surrounding text
 
 #### Who's affected by the problem?
 
 * Screen reader users
 * Users with colour blindness
 * Low-vision users who use high contrast modes
+* Users with monochromatic displays or displays with inaccurate colour reproduction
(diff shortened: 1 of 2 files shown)


accessibility/common-remediations/colours.md
```diff
@@ -17,12 +17,14 @@ Examples may include (but are not limited to):
 * Changing the background colour of an input field to red to indicate an error without also providing an error message
 * Changing the colour of an icon or text string to indicate some kind of status
 * Using a graph or a chart where colour is the only way to indicate different aspects of the data
+* Removing underlines from hyperlinks, leaving just colour to differentiate them from surrounding text
```
hollsk, line 20: ahhh fab thanks so much 🙏 I'd forgotten colour alone was included in this set of docs too. Could you explicitly add an "underline your links, dang it" resolution to the "How do I fix it?" section in this one also, please?

It'll be a bit redundant with it here and in the checklist, but if we ever pull the common remediations out into a standalone set of documents (i.e. into a bigger accessibility guide that's not crammed into the Playbook) then it'll be worth having it directly stated here as well.

morgaan (dismissed):
Fabulous, this contribution warms my heart! Thank you!

We are migrating away from being sociopaths 🥳 
3 persons wearing black t-shirts with white font color where it's written "underline your f*cking links you sociopaths"

accessibility/common-remediations/colours.md
Heydon: @hollsk Ah nice, yeah, will do.

hollsk (approved):
🚀

PR #438, opened 2022-10-10
Direct people to skip link component in Elements

hollsk:
The advice in the checklist is a little out of date - but the component in Elements is not. Instead of replicating the documentation from Elements, I've just removed the outdated advice and directed people to the design system instead, so we're not maintaining the info in two separate locations.

### DIFF
--- a/accessibility/accessibility-checklist.md
+++ b/accessibility/accessibility-checklist.md
@@ -48,6 +48,10 @@ To help them out you can add a "Skip Navigation" link to the top of each page wh
 
 The Elements Design System includes [https://elements.springernature.com/springernature/components/global-skip-link](a skip link implementation) that you can use in your own project(s). 
 
+Resources:
+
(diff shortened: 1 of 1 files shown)


accessibility/accessibility-checklist.md
```diff
@@ -46,25 +46,7 @@ Not all users are able to easily navigate by landmarks. Multiple navigation link
...
-<p>This is the first paragraph</p>
-```
-
-We add `tabindex="-1"` to the targeted element so that it can be programmatically focussed when skipped to.
-
-You should supply CSS that hides the Skip Navigation _for sighted users only_, and makes the link highly visible when it receives focus via tabbing.
-
-Resources:
-
-* [Article: WebAIM's overview of Skip Navigation techniques](https://webaim.org/techniques/skipnav/)
```
sturobson, line 66: would it be beneficial to keep this link in for 'further reading'?

hollsk: Yeah, that's a good idea. Got a bit delete key-happy :D

PR #440, opened 2022-11-22
change developing countries to economies

emsel5201:
a contributor stated 

'replacing “people in developing countries” in “What is accessibility?” with “people with a slow internet connection” or something similar. The term “developing country” is too vague and makes wide assumptions about most of the world (including places like India and China), having been retired by some institutions such as the World Bank in 2016.'

fixes 'list of accessibility-related resources' link


josebolos (approved):
Good catch! :+1:

PR #442, opened 2022-12-12
Change character limit info

emsel5201:
Remove character limit as no longer true.

### DIFF
--- a/accessibility/alt-text.md
+++ b/accessibility/alt-text.md
@@ -55,7 +55,7 @@ Remember what you're trying to achieve - someone who can't see the image needs t
 
 * Don’t say ‘Image of…’ or ‘Picture of…’ etc.  
 * Don’t need to repeat information already given in the text on the page.  
-* Keep it short – 125 characters or less if possible (some screen-readers cut off alt text after this point).
+* Be as succinct as possible but include details as needed. (There is no character limit despite some guidance stating otherwise, [a screen reader won't truncate alt text.](https://yatil.net/blog/there-is-no-character-limit-for-alt-text)) 
 * Copyright info, image source, or other extra information should go in the caption, not the alt text.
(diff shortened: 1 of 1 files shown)


accessibility/alt-text.md
```diff
@@ -55,7 +55,7 @@ Remember what you're trying to achieve - someone who can't see the image needs t
 
 * Don’t say ‘Image of…’ or ‘Picture of…’ etc.  
 * Don’t need to repeat information already given in the text on the page.  
-* Keep it short – 125 characters or less if possible (some screen-readers cut off alt text after this point).
+* Be as succinct as possible but include details as needed. (There is no character limit despite some guidance stating otherwise, [a screen reader won't truncate alt text.](https://yatil.net/blog/there-is-no-character-limit-for-alt-text))
```
hollsk, line 58:
```suggestion
* Be as succinct as possible but include details as needed. There is no hard character limit despite some guidance stating otherwise, [a screen reader won't truncate alt text.](https://yatil.net/blog/there-is-no-character-limit-for-alt-text)
```

Super, thanks!! I think longer term, we're probably going to want to add a full section on alt text length, as it's hard to condense all of the advice into a single sentence. 

We'll need to cover that alt text should be as long as you need it to be to express the concept of what's being shown in the images, but that extra long alt text should be avoided for user experience reasons (and what those UX reasons are); that the length you can get away with probably depends on the content (e.g. NASA's one is long but fine because you don't need to read-and-reread to take it in, but complete alt text for very complex scientific content is maybe better expressed in visible text where you've got full HTML support, etc.); probably other things too. Would be good to get a bit of research going so we can expand this with references.

hollsk: (not in this PR, though, in another one, another day 😅 )

hollsk (approved):
👍

PR #443, opened 2023-01-10
Adds guidance for our preferred use of media queries in Sass

benjclark:


sturobson (approved):

benjclark:
>do we need to find a general consensus amongst those writing (S)CSS that this is the way?

Thanks. Yes, in my view, this PRs provides a good way for us to discuss these things and arrive at our consensus. Also, because this is a public repo, having the discussions in the PR seems best to make them visible to all contributors.

>it would be good to see if it can be added to the (currently wip(?)) Sass lint > stylelint changes
Yes, we could add this as a point to remember to implement as part of that work 👍

PR #444, opened 2023-02-06
Clean up browser support page

josebolos:
* Remove some leftover references to Graded Browser support that don't make much sense anymore:
  * Fixing bugs with "high priority" on Enhanced browsers (as we try to fix all known bugs as soon as we can)
  * References to "old versions of evergreen browsers" as in the grey area (as all those versions are now all Core)
* Move the .browserslist bit to the implementation section, and the Caveats out of it
* Update Table of Contents to reflect new structure


hollsk (approved):
A+ PR, would review again 👍

PR #447, opened 2023-04-03
Delete questionable CSS practice

RSeidelsohn:
In the same [document](https://github.com/springernature/frontend-playbook/blob/8d09fc5a25a738c75d53aeba0f9d28f0db8ff69d/css/house-style.md), we explain that we should prefer a class over HTML tags (with some exceptions - which are not explained) and then show how we nest HTML tags, where still classes would do.

Also, the example to me doesn't make any sense - in the proposed case when items semantically must sit within other items, then why nest them in CSS rules? If a `<td>` must sit within a `<tr>` and that within a `<table>`, then why not directly style the `<td>`?

If these rules make sense, then I am not getting it and that _might_ be an indicator that some more detailed documentation could help here.

So I want to put this section up for discussion.

The only possibly useful nesting here I can see would be the `ul > li` one, as it explicitly differs from an `<li>` nested within an `<ol>`, even though I still would propose that a class would make more sense here, like for example `.c-inline-list > li`.

### DIFF
--- a/css/house-style.md
+++ b/css/house-style.md
@@ -122,33 +122,6 @@ span {
 }
 ```
 
-**Items that semantically MUST sit within other items**. We do this:
-```scss
-ul {
-    > li {
-        display: inline;
-    }
-}
-
-dl {
-    > dt,
-    > dd {
-        display: inline-block;
-    }
-}
-
-table {
-    > tr {
-        > th {
-            background: #ccc;
-        }
-        > td {
-            background: #eee;
-        }
-    }
-}
-```
-
 **Javascript-only style enhancements**. We do this:
 ```scss
 .c-class {


benjclark:
I agree that the example of nesting with nested HTML tags is in conflict with the guidance about preferring classes. Good idea to remove this to prevent confusion.

benjclark (approved):

alexkilgour:
Regarding the bit about "no exceptions", I think it's possible that the case of semantically nested items _is_ the expectation and that just isn't explained very well. The two arguments I can think of for keeping this advice would be:

1. It could make your CSS more readable than splitting into separate classes
2. Fewer classes in the HTML, again improving legibility 

Not saying I disagree with removing this block, but I'm trying to consider why we added it in the first place!

RSeidelsohn:
> 2. Fewer classes in the HTML, again improving legibility

While I agree that fewer classes in HTML are a good thing, I would still vote for default style changes better being denoted by (speaking) CSS classes rather than by styling HTML elements. At least, I would suggest to not promote styling HMTL elements directly,  but of course this is not the case here.

RSeidelsohn:
BTW, this case is a great example for learning why prose in commit comments is a great thing - not that I would lead by example here - but nothing beats a commit message that explains, several years later, why a commit has been deemed necessary, even if at that time it was obvious for everyone involved.

PR #448, opened 2023-05-10
Very minor renaming of git.md section

dotcode:
"Repositories" was too broad …

### DIFF
--- a/git/git.md
+++ b/git/git.md
@@ -1,6 +1,6 @@
 # Git
 
-- [Repository names](#repositories)
+- [Repository names](#repository-names)
 - [Working with branches](#working-with-branches)
 - [Commit messages](#commit-messages)
 - [Pull requests](#pull-requests)
--- a/git/git.md
+++ b/git/git.md
@@ -1,12 +1,12 @@
 # Git
 
-- [Repositories](#repositories)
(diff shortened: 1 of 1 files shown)


git/git.md
```diff
@@ -1,12 +1,12 @@
 # Git
 
-- [Repositories](#repositories)
+- [Repository names](#repositories)
```
sturobson, line 3: as this is linking to the ID of the H2 that's created - I think it needs to be

`[Repository names](#repositories-names)`

git/git.md
```diff
@@ -1,12 +1,12 @@
...
 - [Working with branches](#working-with-branches)
 - [Commit messages](#commit-messages)
 - [Pull requests](#pull-requests)
 - [Versioning](#versioning)
 
-## Repositories
+## Repository names
```
sturobson, line 9: the only (very minor) problem that I can see arising is if `file/to/git#repositories` is linked from elsewhere - but the consumer would still end up on the page (at the top, rather than the linked section) … as the page has the toc - they should be able to 'find their way' :)

dotcode: Ooh good point, ta … I did a search, apparently it's not linked directly from anywhere 👍

PR #450, opened 2023-07-06
Removes (editors draft) from WCAG section

emsel5201:
Documents are no longer editors drafts and are published documents


hollsk (approved):
🎉

PR #451, opened 2023-07-06
Changes to the why do we care about a11y section

emsel5201:
Giving more weight to serving diverse users over legalities


jpw (approved):
Great, thank you!

dotcode (approved):
Nice - thank you!

PR #452, opened 2023-07-17
Fixes links in Semantics and Structures

emsel5201:
Fixes elements link for skip link implementation, removes dead link to Samurai Errata


hollsk (approved):
Samurai errata website is dead now - looks like it was taken down in 2020. 

I'd normally recommend keeping archival versions of dead links in the Wayback Machine, but in this case... ehh. WCAG 2.x has been out for a long time, and I'm fairly sure everyone who once used the Samurai errata is no longer doing so (happy to be corrected on this, but I've not seen anyone using or referring to it in years and years), so I'm happy to let this one go peacefully to its grave. 

Thanks Emily!

PR #455, opened 2023-09-20
Add section about indentation

RSeidelsohn:
As I did not find any notion of why we promote the use of `tab`s vs. `space`s for source code indentation, I did some discovery work and added a section to our `practices/house-style.md` document.


css/house-style.md
```diff
@@ -122,33 +122,6 @@ span {
 }
 ```
 
-**Items that semantically MUST sit within other items**. We do this:
```
jpw, line 125: why has this been deleted?

jpw (changes requested):
i think the addition looks great, thanks!
Please see comment about some code removal though.

css/house-style.md
RSeidelsohn: Hm, this somehow ended up in this PR, although it comes from [another PR that already has been approved](https://github.com/springernature/frontend-playbook/pull/447).
In that one, I was deleting a section that propoesed something that according to our Playbook we should not do (styling HTML elements rather than classes).

practices/house-style.md
```diff
@@ -21,6 +21,12 @@ Read this excellent [article by Nicholas C. Zakas](https://www.smashingmagazine.
 
 Read the [markup guide](../markup/house-style.md), the [JavaScript guide](../javascript/house-style.md) and the [CSS guide](../css/house-style.md) to understand our house style in these technologies. You should also familiarise yourself with our [Git](../git/git.md) strategies. The team you're deployed to may also have its own strategies in addition to these.
 
+### Indentation
+
+With regards to accessibility, we favour using `tab`s for indentation, rather than `space`s. Using `tab`s, every person can define their preferred _render width_ as they like. This also means, that we discourage setting any `tab width` settings in any shared configuration file, as that's a display thing, not a code thing.
```
jon-stevens, line 26:
```suggestion
With regards to accessibility, we favour using `tab`s for indentation, rather than `space`s. Using `tab`s, every person can define their preferred _render width_ as they like. This also means that we discourage `tab width` settings in any shared configuration files, as that's a display thing, not a code thing.
```

css/house-style.md
jpw: Can we remove from this PR please? Does this branch need updating?

RSeidelsohn: I was hoping to get around this, but I agree that this is not nice, so I'll update the branch. Thanks for pointing that out.

jpw: Think this still needs doing? (sorry)

RSeidelsohn: Sorry for leaving it untouched that long. I now reverted the change and hope that it will find it's way from the even older PR...

RSeidelsohn: No wait, it already has been merged. I now merged the current master into this PR in order to prevent my revert from undoing the change that already landed in master.

jpw (approved):

PR #457, opened 2023-10-26
Update secure markup page

jpw:
This PR intends to:
 - removes references to the `typemustmatch` attribute, which was never adopted beyond Firefox and is since deprecated.
 - use more sensitive language.
 - update some links.
 - update advice for the `object` element.

This page is a bit of a weird bag of HTML tips, without much structure and not very useful. I think the playbook would benefit from a more comprehensive introduction to frontend-related security topics, which I do have plans for...


RSeidelsohn (approved):
Nothing to complain about. This is an improvement of our documentation.

jpw:
@RSeidelsohn thank you

PR #458, opened 2023-10-31
Small clarifications to language and concepts

hollsk:
* change all instances of "compliance" to "conformance" ([they're not the same thing](https://www.kellertechnology.com/blog/compliance-vs-conformance-understanding-the-relationship-to-standard/))
* changed a "verification" to "validation" ([they're not the same thing either](https://www.geeksforgeeks.org/differences-between-verification-and-validation/))
* make it clear that you're not to rely on _automated testing_ alone, and not that you're not to rely on the contents of _this page_ alone (original seemed ambiguous to me)
* clarify that Microsoft's Accessibility Insights is not a silver bullet for lurking accessibility problem werewolves
* clarify that screen reader users use all sorts of screen readers, so we don't just pick one and target it like the bad old days of Netscape Navigator et al

### DIFF
--- a/accessibility/tools.md
+++ b/accessibility/tools.md
@@ -12,11 +12,11 @@
 
 ## Automated testing
 
-Automated accessibility testing can be used to detect the most common accessibility problems. This is only the first step in the process; you MUST NOT rely on automated testing as your sole method of verification! None of these tools on their own will catch every error. Even by combining all of them, it's still possible to produce an inaccessible webpage. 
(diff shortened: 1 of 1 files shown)


jpw (approved):
LGTM, thanks :)

RSeidelsohn (commented):
Improving is an ongoing and never-ending process.
These changes really improve the clarity of the Playbook as well as add more useful details.
Thanks a lot! I can't find any other comments for this PR, but I am unfortunately also not allowed to approve it, sorry for that.

PR #460, opened 2023-11-10
Update Jake Archibald details

asfafk:


dotcode:
looking at LinkedIn I can see that @jakearchibald is now Principal Engineer at Shopify. So yep, looks good to me!

dotcode (approved):

PR #462, opened 2023-11-28
adds images of text guidance to alt text

emsel5201:
provides guidance on writing alt text for images of text

### DIFF
--- a/accessibility/alt-text.md
+++ b/accessibility/alt-text.md
@@ -19,6 +19,9 @@
   - [Headshots](#headshots)
     - [Bad alt text](#bad-alt-text-3)
     - [Good alt text](#good-alt-text-3)
+  - [Images of text](#images-of-text)
+    - [Bad alt text](#bad-alt-text-4)
+    - [Good alt text](#good-alt-text-4)
 
 
 ## What is alt text? 
@@ -167,4 +170,28 @@ However, you MUST be careful with this - not everyone is comfortable with having
 * [Writing great alt text: Emotion matters](https://jakearchibald.com/2021/great-alt-text/)
 * [Thoughts on skin tone and text descriptions](https://tink.uk/thoughts-on-skin-tone-and-text-descriptions.md-notes-on-synthetic-speech/), by Léonie Watson
 
+### Images of text 
 
+<img width="625" alt="screenshot of social media post of an image of text" src="https://github.com/springernature/frontend-playbook/assets/69847167/83772c8e-6b61-4f5a-8970-fc91be79f84d">
+
+In a Springer Nature social media post, this image of text and book covers was accompanied with the copy: “Are you a horror enthusiast interested in finding out more about vampires in film? Then this book is a must-read for you! From the first ever filmic female vampire to a Dracula look-alike and cinema’s original vampire, this book discusses them all. https://bit.ly/3SnsTrN” 
+
+Images of text are often used on social media and in advertisements for visual appeal and to convey information quickly. But images are one of the least accessible ways of presenting text - if a user needs to magnify or zoom their screen, the images can become distorted or pixellated, rendering the text unreadable. If the image is complex, like this one, then it’s much harder to write sufficient alt text for it. You should avoid publishing images of text as much as you can. If it’s unavoidable, use as little text as possible, and don’t make the text in the image your only call to action.
+
+This example is a complex graphic with a lot happening in it, and it’s easier to start with what you shouldn’t do with an image like this:
+
+#### Bad alt text
+
+* “text and book covers”
+* “Vampires on the silent screen” 
+* “Image of book cover”
+
+#### Good alt text
+
+“Cover image of Vampires on the silent screen: Cinema’s First Age of Vampires 1897-1922, by David Annwn Jones, published by Palgrave Macmillan for Palgrave Gothic. The cover image features a monochrome illustration of Max Schreck as Count Orlok in the 1922 film Nosferatu: A Symphony of Horror. The book provides a unique insight into the iconography of vampires in silent cinema, presenting a detailed academic yet accessible discussion of the films and their sources, and re-configuring the history of horror cinema.”
+
+This alt text does the following things:
+
+1. Incorporates, verbatim, the text to the left of the cover image
+2. Details the title of the book, the name of the book’s author, the publishing house, and the publication series the book belongs to - all visible on the cover image
+3. Describes the cover illustration, which is of a highly-recognisable film character that’s directly relevant to the book’s subject matter, and that the target audience for this book would be expected to be familiar with


accessibility/alt-text.md
```diff
@@ -167,4 +170,28 @@ However, you MUST be careful with this - not everyone is comfortable with having
...
 
+### Images of text 
 
+<img width="625" alt="screenshot of social media post of an image of text" src="https://github.com/springernature/frontend-playbook/assets/69847167/83772c8e-6b61-4f5a-8970-fc91be79f84d">
+
+In a Springer Nature social media post, this image of text and book covers was accompanied with the copy: “Are you a horror enthusiast interested in finding out more about vampires in film? Then this book is a must-read for you! From the first ever filmic female vampire to a Dracula look-alike and cinema’s original vampire, this book discusses them all. https://bit.ly/3SnsTrN” 
+
+Images of text are often used on social media and in advertisements for visual appeal and to convey information quickly. But images are one of the least accessible ways of presenting text - if a user needs to magnify or zoom their screen, the images can become distorted or pixellated, rendering the text unreadable. If the image is complex, like this one, then it’s much harder to write sufficient alt text for it. You should avoid publishing images of text as much as you can. If it’s unavoidable, use as little text as possible, and don’t make the text in the image your only call to action.
```
jpw, line 179:
```suggestion
Images of text are often used on social media and in advertisements for visual appeal, brand consistency and to convey information quickly. But images are one of the least accessible ways of presenting text - if a user needs to magnify or zoom their screen, the images can become distorted or pixellated, rendering the text unreadable. If the image is complex, like this one, then it’s much harder to write sufficient alt text for it. You should avoid publishing images of text as much as you can. If it’s unavoidable, use as little text as possible, and don’t make the text in the image your only call to action.
```

accessibility/alt-text.md
```diff
@@ -167,4 +170,28 @@ However, you MUST be careful with this - not everyone is comfortable with having
...
+
+This example is a complex graphic with a lot happening in it, and it’s easier to start with what you shouldn’t do with an image like this:
+
+#### Bad alt text
+
+* “text and book covers”
+* “Vampires on the silent screen” 
+* “Image of book cover”
```
jpw, line 187: Perhaps some explanation as to why they are "bad" would be helpful?

accessibility/alt-text.md
```diff
@@ -167,4 +170,28 @@ However, you MUST be careful with this - not everyone is comfortable with having
...
+
+#### Good alt text
+
+“Cover image of Vampires on the silent screen: Cinema’s First Age of Vampires 1897-1922, by David Annwn Jones, published by Palgrave Macmillan for Palgrave Gothic. The cover image features a monochrome illustration of Max Schreck as Count Orlok in the 1922 film Nosferatu: A Symphony of Horror. The book provides a unique insight into the iconography of vampires in silent cinema, presenting a detailed academic yet accessible discussion of the films and their sources, and re-configuring the history of horror cinema.”
+
+This alt text does the following things:
+
+1. Incorporates, verbatim, the text to the left of the cover image
+2. Details the title of the book, the name of the book’s author, the publishing house, and the publication series the book belongs to - all visible on the cover image
+3. Describes the cover illustration, which is of a highly-recognisable film character that’s directly relevant to the book’s subject matter, and that the target audience for this book would be expected to be familiar with
```
jpw, line 197:
```suggestion
3. Describes the cover illustration; which is of a highly-recognisable film character that’s directly relevant to the book’s subject matter, and that the target audience for this book would be expected to be familiar with
```

jpw (changes requested):
Really useful addition and a great example, thank you.

Something perhaps for a new PR, not sure what we think about this... I think it might be a bit mean to call out actual examples and describe them as "bad". I didn't suggest changes on this point in the PR as I see it's a pattern used throughout this document.

In the future, perhaps we could amend to "Problematic" and "Better" instead of "Good" and "Bad"?

accessibility/alt-text.md
```diff
@@ -167,4 +170,28 @@ However, you MUST be careful with this - not everyone is comfortable with having
 * [Writing great alt text: Emotion matters](https://jakearchibald.com/2021/great-alt-text/)
 * [Thoughts on skin tone and text descriptions](https://tink.uk/thoughts-on-skin-tone-and-text-descriptions.md-notes-on-synthetic-speech/), by Léonie Watson
 
+### Images of text 
 
+<img width="625" alt="screenshot of social media post of an image of text" src="https://github.com/springernature/frontend-playbook/assets/69847167/83772c8e-6b61-4f5a-8970-fc91be79f84d">
```
hollsk, line 175: alt text examples go here: https://github.com/springernature/frontend-playbook/tree/main/accessibility/images/alt-examples (there's an upload button top right on that page, if you're using the web interface on Github), then your src will be `src="images/alt-examples/orlok.png"` (or whatever file name you're using)

hollsk (changes requested):

accessibility/alt-text.md
emsel5201: I think the guiding principles section covers why this would be insufficient information

A few guiding principles

- Don’t say ‘Image of…’ or ‘Picture of…’ etc. (It is nearly always wrong to say an "Image of", because all images are … images. It is extraneous information that the screenreader announces. However, it can help to know the type of image - i.e. is it a photo, or a screenshot, or a cartoon, etc, depending on the surrounding content and context.)
- Don’t need to repeat information already given in the text on the page.
- Be as succinct as possible but include details as needed. There is no hard character limit despite some guidance stating otherwise, a screen reader won't truncate alt text.
- Copyright info, image source, or other extra information should go in the caption, not the alt text.
- Context is important e.g. it may be appropriate to note if an image is a photo or an artwork, whether a person is old or young, if the sky is stormy or clear, if that is relevant to the content.
- Think about how you would describe the image to a person who can’t see it.


Do you think so or is it worth adding more?

emsel5201:
> Really useful addition and a great example, thank you.
> 
> Something perhaps for a new PR, not sure what we think about this... I think it might be a bit mean to call out actual examples and describe them as "bad". I didn't suggest changes on this point in the PR as I see it's a pattern used throughout this document.
> 
> In the future, perhaps we could amend to "Problematic" and "Better" instead of "Good" and "Bad"?

I will change these in a separate PR

accessibility/alt-text.md
emsel5201: will change on another PR

accessibility/alt-text.md
jpw: The other examples have an explanation underneath describing _why_ these alt texts are insufficient, whereas these do not.

accessibility/alt-text.md
```diff
@@ -167,4 +169,30 @@ However, you MUST be careful with this - not everyone is comfortable with having
...
+
+This example is a complex graphic with a lot happening in it, and it’s easier to start with what you shouldn’t do with an image like this:
+
+#### Insufficient alt text
+
+* “text and book covers”
+* “Vampires on the silent screen” 
+* “Image of book cover”
+  
+the purpose of the image is to promote the book. None of these descriptions tell the user anything about the book, so there’s no reason why a person who can’t see the image would want to buy it.
```
hollsk, line 188:
```suggestion
The purpose of this image is to promote the book. None of these descriptions tell the user anything about the book, so there’s no reason why a person who can’t see the image would want to buy it. 
```

hollsk (approved):
Thanks! Approving now so I'm not blocking your update, assuming jpw is happy with the explainer for the insufficient examples 👍

PR #465, opened 2023-12-13
add some guidance on alt text length

hollsk:
Many people have asked us about alt text character limits. There are none, but there are certainly UX considerations on how much you stuff in there. This PR attempts to finally address people's questions with proper Playbook guidance. 

I've moved the original "no hard character limit" clause from the bulletpoint under "a few guiding principles" and incorporated it into a fuller section further up. 

At a later date I'll think about guidance for full text transcripts (but that is very much not a job for my last hour before finishing for the year 🙃)


accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+Many HTML element nodes can contain attributes. A link may use a `title` attribute, like this ([though I do not recommend it](https://github.com/springernature/frontend-playbook/blob/main/accessibility/common-remediations/html-attributes.md)): `<a href="/" title="don't do this">my cool website address</a>`. The alt attribute is also of this type: `<img alt="my cool alt text" src="...">`. Attributes can't be interacted with in the same way as the rest of the page - you can't move forwards or backwards through the text, you can't read the text at the character level, and you can't pause the screen reader and restart later where you left off.
```
emsel5201, line 64: might (though I do not recommend it) read better after the example, for me it was a bit difficult to read.

jpw: I was wondering why we are promoting an a11y anti-pattern "even though we don't recommend it", but I suppose if its a widespread practice then it does good to cast it more into disrepute.

accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+If the information is complex (and much of Springer Nature's scientific output is highly complex), the cognitive load on the screen reader user is higher, as they have more information to process and recall. For (many, not all) sighted users, it's a simple task to scan your eyes back and forth to read a complex sentence or sentence fragment again. It's impossible for a blind screen reader to do this with alt attributes, and this is the reason that alt text must be succinct.
```
jpw, line 68:
```suggestion
If the information is complex (and much of Springer Nature's scientific output is highly complex), the cognitive load on the screen reader user is higher, as they have more information to process and recall. For many (but not all) sighted users, it's a simple task to scan your eyes back and forth to read a complex sentence or sentence fragment again. It's impossible for a blind screen reader to do this with alt attributes, and this is the reason that alt text must be succinct. 
```

accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+If the information is complex (and much of Springer Nature's scientific output is highly complex), the cognitive load on the screen reader user is higher, as they have more information to process and recall. For (many, not all) sighted users, it's a simple task to scan your eyes back and forth to read a complex sentence or sentence fragment again. It's impossible for a blind screen reader to do this with alt attributes, and this is the reason that alt text must be succinct. 
+
+Don't make it so succinct that the information in the image is lost, but do think about the cognitive burden of alt text as a whole, and consider if it's really alt text in the HTML that you need, or if your users would be better served with a full text transcript that they can directly interact with instead.
```
jpw, line 70:
```suggestion
Don't make it _so_ succinct that the information in the image is lost, but _do_ think about the cognitive burden of alt text as a whole; and consider if it's really alt text in the HTML that you need, or if your users would be better served with a full text transcript that they can directly interact with instead. 
```

jpw (changes requested):
A worthy addition! Thank you. However I do have quibbles like tribbles.

accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+You MUST keep alt text in HTML as succinct as you reasonably can while still expressing all the information in the image. To understand why, you need to understand how screen readers parse attributes like `alt`. 
+
+### How do screen readers render alt text? 
+
+Screen readers look at all of the HTML markup on a page, then announce the contents to the user in different ways depending on what kind of element it's describing. For example a link might be announced as something like "Link: my cool website address". A navigation menu might be announced like "Navigation: breadcrumb". These are examples of HTML element nodes. Screen reader users can command their screen reader to move around the page and describe different things at will. They can move forwards or backwards through text, read text at the character level (for example to confirm spelling), and pause and restart later, among other useful operations.
```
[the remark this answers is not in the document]
hollsk:
```suggestion
Screen readers look at all of the HTML markup on a page, then announce the contents to the user in different ways depending on what kind of element it's describing. For example a link might be announced as something like "Link: my cool website address". A navigation menu might be announced like "Navigation: breadcrumb". These elements are examples of HTML nodes. Screen reader users can command their screen reader to move around the page and describe different things at will. They can move forwards or backwards through text, read text at the character level (for example to confirm spelling), and pause and restart later, among other useful operations. 
```

better?

accessibility/alt-text.md
hollsk: @jpw 2 reasons: 

1. There just aren't that many attributes to choose from that are exposed as announcements to screen readers. You're not going to get a data attribute exposed, for example, or an href, or size information etc, so not much point invoking them here as examples. I really only have a couple of things I can choose from - `title`, which I did use, or `aria-label` / `aria-description`, both of which would suffer similar problems and need to be even more heavily caveated, and that we don't currently have Playbook guidance written up on (it's on my todo list). If you can think of another possibility I'd be very glad to hear it, though.
2. You're correct that `title` abuse is a VERY widespread practice, and I take every opportunity I get to say it shouldn't be done 😅 



@emsel5201 How do you feel about this as a rewrite? 

```suggestion
Many HTML element nodes can contain attributes. [A link could (but shouldn't) use a `title` attribute](https://github.com/springernature/frontend-playbook/blob/main/accessibility/common-remediations/html-attributes.md)), like this: `<a href="/" title="don't do this">my cool website address</a>`. The alt attribute is also of this type: `<img alt="my cool alt text" src="...">`. Attributes can't be interacted with in the same way as the rest of the page - you can't move forwards or backwards through the text, you can't read the text at the character level, and you can't pause the screen reader and restart later where you left off. 
```

emsel5201: Yes, I like the rewrite!

accessibility/alt-text.md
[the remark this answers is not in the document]
jpw: Yes thank you, but perhaps "These elements are examples of HTML nodes" is redundant now? Not sure what you are trying to achieve by pointing this out.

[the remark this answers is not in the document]
jpw: Looking again in the wider context of the paragraph, I can see now what you are intending, thanks :)

jpw (dismissed):

accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+If the information is complex (and much of Springer Nature's scientific output is highly complex), the cognitive load on the screen reader user is higher, as they have more information to process and recall. For many (but not all) sighted users, it's a simple task to scan your eyes back and forth to read a complex sentence or sentence fragment again. It's impossible for a blind screen reader to do this with alt attributes, and this is the reason that alt text must be succinct. 
+
+Don't make it _so_ succinct that the information in the image is lost, but _do_ think about the cognitive burden of alt text as a whole; and consider if it's really alt text in the HTML that you need, or if your users would be better served with a full text transcript that they can directly interact with instead.
```
dotcode, line 70: Given that transcripts are most commonly associated with audio/video, would it be better to instead reference something like a caption?

hollsk: hm. it's an interesting question. _We_ often leverage the caption for this on our journal content, but it's not going to be appropriate for every use case - when I wrote this, I was mainly thinking of [full text transcripts like this one from the CDC](https://www.cdc.gov/ncbddd/disabilityandhealth/infographic-disability-impacts-all.html) (scroll down to the bottom of the page to see it), which isn't in a caption and doesn't really need to be. Also thinking of stuff like transcripts that live under progressive disclosure widgets (could use `details/summary`, could just use a load of divs with an aria-suguared button) that are used on a lot of our immersive articles. 

I'm hesitant to specify a particular element here, I guess, as it's unnecessarily prescriptive. What I do want to do is add another page to this guide that just covers different options for text transcripts (and then I'll link to it here). Does that assuage your concerns? 🤞

accessibility/alt-text.md
```diff
@@ -47,6 +49,26 @@ No. If an image has no meaning, or if it's providing information that's redundan
...
+
+Imagine you're sitting at a bus stop, and the bus company has added a digital marquee to the shelter to give you important information. You can't control the marquee at all, you can only look at it, passively, on its own schedule. If you're momentarily distracted, causing you to miss the information you needed about your bus, the only thing you can do is wait for it to return to the start and then (eventually) get to the part you were interested in. Tiresome, right? This is similar to the experience of using alt text with a screen reader. Although the screen reader user can go back to start whenever they want, if your alt text is 1000 characters long and the information they need to hear again starts after 850 words, this person is going to be waiting for a very long time.
```
tulipdexter, line 66:
```suggestion
Imagine you're sitting at a bus stop, and the bus company has added a digital marquee to the shelter to give you important information. You can't control the marquee at all, you can only look at it, passively, on its own schedule. If you're momentarily distracted, causing you to miss the information you needed about your bus, the only thing you can do is wait for it to return to the start and then (eventually) get to the part you were interested in. Tiresome, right? This is similar to the experience of using alt text with a screen reader. Although the screen reader user can go back to the start whenever they want, if your alt text is 1000 characters long and the information they need to hear again starts after 850 words, this person is going to be waiting for a very long time. 
```

ruthmoog (approved):
This is nice and relatable guidance 👍

hollsk:
Thanks everyone! 🎉 ❤️

PR #468, opened 2024-01-19
Add link to the eight landmark roles 

ricky-lau:
Replaced http://www.html5accessibility.com/#sections which was referencing HTML sectioning elements instead of landmarks with https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles#3._landmark_roles.


hollsk (approved):
Whoops sorry Ricky - I thought I'd approved this already but I think I must have just vividly imagined approving it instead of actually doing it 🙈 

Thanks for clarifying this one!

PR #472, opened 2024-03-06
Clarify practices regarding browser polyfills and third party CDNs

josebolos:
- Adds link explaining what a browser polyfill is.
- Adds the data protection risk to the "use of third party CDN" risks.
- Makes it clearer that using third party CDNs is not an acceptable practice, ever, both in the "static resources" page and in the "polyfill" section of the javascript house style page.
- Also minor whitespace adjustments and updating TOCs for consistency.

### DIFF
--- a/javascript/house-style.md
+++ b/javascript/house-style.md
@@ -2,51 +2,53 @@
 
 This document outlines the way we write JavaScript. It's a living style guide – it will grow as our practices do.
 
-* [General principles](#general-principles)
-  * [Progressively enhance](#progressively-enhance)
-  * [Code for humans](#code-for-humans)
-  * [Modules over monoliths](#modules-over-monoliths)
-  * [Keep it simple](#keep-it-simple)
-  * [Performant code](#performant-code)
-* [Code style](#code-style)
-  * [Linting](#linting)
-  * [Comments](#comments)
-  * [Asynchronicity](#asynchronicity)
-    * [Things to consider](#things-to-consider)
-    * [Tips](#tips)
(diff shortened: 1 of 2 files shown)


benjclark:
I wonder if we should suggest that any NPM dependency used to load polyfills should be locked down to an exact version in a package.json / package-lock.json, and that version range syntax should not be used. To mitigate risk of an NPM dependency getting compromised and publishing code that we import and load in production. But I guess that applies to any  NPM dependency used in production code.

josebolos:
@benjclark Yeha, good point. Interestingly, we also suggest using [SRI](https://github.com/springernature/frontend-playbook/blob/main/security/secure-markup.md#use-subresource-integrity) which, if my understanding is correct, could also help with these issues. I don't think we use it extensively, though.

josebolos:
Hi @benjclark (and everyone else) - thinking more about this, it looks like the dependency version locking would be a much bigger (and very worthwhile) conversation to have, espcially in light of security news in the last few months/years.

Given most of that work should probably end up in the ["Managing Node Projects" doc](https://github.com/springernature/frontend-playbook/blob/4a4d8033dd8e035de1e67b777c794306e75139ea/practices/managing-node-projects.md), and that this PR was focused on removing specific third party providers from the JavaScript house style doc, would you be ok with this being merged as it is? Dependency management could probably be discussed in the Open Space instead.

benjclark:
@josebolos No objection from me about merging this as is, makes sense 👍

benjclark (approved):

PR #476, opened 2024-03-19
Use Wayback Machine for dead link

### LINKED ISSUE #475: simplyaccessible.com no longer exists
https://github.com/springernature/frontend-playbook/blob/68ae629216510730139866f39ed36fa27360e477/practices/progressive-enhancement.md?plain=1#L24

This website appears to no longer exist, we should replace this with a different reference.

hollsk:
Resolves #475 

It's still a good quote, and it's still a good post, and the Internet Archive has got our backs. Which reminds me, I was going to donate to them ahead of the elections in 2024. Thanks for jogging my memory!


dotcode (approved):
lgtm, thanks :)

PR #477, opened 2024-04-23
Rearrange the accessibility content so it's easier to read

hollsk:
## Now what is your problem, hollsk
Github recently deployed a change that makes individual pages on github look like they're being served directly out of an operating system file system thingy. We're now in a situation where the README is confronting our poor users with three separate tables of contents and no actual, y'know, content. Look, here's a screenshot:

an image of the Github single page interface that has a list of files in the left hand sidebar, the same list of files at the top of the main content area, and the same list of files as the content itself

## How awful
I really agree! I share these pages often with human people who aren't developers, so I want this page to look as if it was written by a human who understands what other humans need and want, which it is not currently achieving. Embarrassing.

So! My changes run thusly:

* remove the old introduction.md file and put its contents into README.md, so users can immediately read some content, instead of looking at the list of files for the third time on the same page load
* rename all the files to give them numbers, so they can be ordered in a sensible way (this is what our old TOC was trying to achieve in the first place)
* renamed some of the files to just make it clearer what they're for - e.g. "developing accessible products" was always for suppliers, but that wasn't apparent from the doc itself
* fix all the links everywhere in the repo so they're not broken


accessibility/README.md
```diff
@@ -1,15 +1,101 @@
...
-* [Tools](tools.md)
-* [Accessibility checklist](accessibility-checklist.md)
-* [Accessibility in emails](accessibility-in-emails.md)
-* [Accessibility in PDFs](accessibility-in-pdfs.md)
-* [Effective colour contrast](effective-colour-contrast.md)
-* [Common remediations](common-remediations/README.md)
-* [Alt text](alt-text.md)
-
-[Main table of contents](../README.md#table-of-contents)
+* [General principles](#general-principles)
```
RSeidelsohn, line 3: Dear Hollie, your changes even improve the readability of this PR! I love it. I just wonder whether it is on purpose that the links to all these documents (of which you changed the names really beautifully!) are now gone?

hollsk: > This is due to the fact that I am used to seeing the codebases on GitHub

Yeah, this is basically where my disconnect is - I share these files with people who aren't used to Github, and who don't have the same "sidebar-blindness" (for want of a better term) than you do. 

The ToC links were always a bit of a bodge. The point of them was so we could put the files into a sensible reading order, as the experience was a bit unguided without it. Nobody was brave enough to rename all the docs in the repo to give them numbers, so we added a ToC in each README instead. But now the bodge is _visibly_ a bodge, especially to people who expect to see content on a webpage, rather than a bunch of files. 

I do take your point, though, and I was in two minds last night about whether I should incorporate the links directly into the intro, with a bit of explanation around them. On the one hand it makes the already fairly long introduction even longer... but on the other hand I think the intro could benefit from more signposting. 

I think what I'll do is add an orientation section near the end, fold the "Resources" and "Tools" headings into it, and explicitly direct people to useful pages. That'll mean it does the same job that the list did, but without the visual display being a literal crime against sense and good taste. Will update the PR shortly, thank you! 🙏

RSeidelsohn: Ok, I understand. I really appreciate your effort and think that this appendix with "useful links" would improve the value of the README.

RSeidelsohn (approved):
I think this is a great improvement.

PR #484, opened 2025-01-06
Adding information about signing and verifying commits to the Git section

dsmr:
As mentioned in the #developers Slack channel and in Hive, we should sign our Git commits. To align with this, I've added instructions for signing and verifying commits on GitHub to the Git section of the playbook.

Signing commits helps ensure that code changes are both authentic and secure, preventing unauthorized modifications and improving traceability. By including this information, the playbook not only promotes best practices but also prepares the team for future security requirements.

### DIFF
--- a/git/git.md
+++ b/git/git.md
@@ -84,3 +84,38 @@ Your work is now merged, good job! Delete your remote branch, and delete your lo
 Our open source projects are versioned with [SemVer](semver.md). Check our [Open Source support](../practices/open-source-support.md) guide for details on how to version and release an update to an open source project.
 
 Closed source projects may use different versioning practices.
+
+## Sign and Verify Your Commits on GitHub
+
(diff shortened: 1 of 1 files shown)


benjclark:
@dsmr Thinking about this a little, I think this information is not relevant for the Playbook. This is because is the Playbook is a public resource for anyone on the internet, whilst this information in your PR is more related to company practices and policy for employees. So I would say lets close this PR and we can think about how best to ensure FEDs are aware of this guidance during onboarding etc.

Hope that makes sense?

dsmr:
@benjclark yeah, you're right, thank you :).

alexkilgour:
I think there is definitely a place for this information (as well as in the onboarding instructions), and that might be in the new confluence when we get it - I know there is a wider discussion around documentation so let's see what happens when that is up and running.

Thank you for being proactive and adding this though, I'm sure we can make use of it at some point 😄

PR #485, opened 2025-02-06
Define "remediations"

dotcode:


jpw:
Thanks @dotcode - I think it would be good to clarify the meaning too.

However IMO having an explanation in parentheses in a title feels awkward; if the word choice could be improved then perhaps we should change the word!?

Perhaps a quick fix in the meantime is to incorporate an explanation in the leading paragraph?

Another minor worry is that this change would break any anchor links pointing to that header, but as it's at the top of the document this would not overly affect inbound links, and we can run a link-checker over internal links.

dotcode:
Ooh that's a good point, thanks! Nicely demonstrating the utility of pull requests, even for the simplest of changes 😂

I'll submit a new PR.

PR #486, opened 2025-02-13
Add clarifications around SVGs in control elements

danielclubb:
- Rename `Decorative SVG inside a control is focusable` section to `SVG inside a control is independently focusable`
- Improve clarity of use case explanations and `we do/we don't do this` sections


hollsk (approved):
Super, thanks so much Daniel!

PR #501, opened 2025-11-14
init suggestion for a SN FED copilot persona for context

Sreilys:
This pr creates a SN FED copilot persona for more context.
Very open to suggestions in how we can expand it even more and give more guidance to ai.

### DIFF
--- a/copilot/README.md
+++ b/copilot/README.md
@@ -1,0 +1,9 @@
+# Copilot
+
+If you must use AI, do so responsibly and in alignment with our frontend playbook.
+This chatmode give Copilot more context about our guidelines and best practices.
+You can copy and paste the following chatmode into your VSCode editor to ensure it follows our guidelines or refer to it by linking it in your project specific custom chatmode.
+
+* [SN FED Chatmode](sn-fed.chatmode.md)
+
+[Main table of contents](../README.md#table-of-contents)
--- a/copilot/README.md
+++ b/copilot/README.md
@@ -1,8 +1,8 @@
 # Copilot
 
 If you must use AI, do so responsibly and in alignment with our frontend playbook.
-This chatmode give Copilot more context about our guidelines and best practices.
-You can copy and paste the following chatmode into your VSCode editor to ensure it follows our guidelines or refer to it by linking it in your project specific custom chatmode.
+This chatmode gives Copilot and other AI tools more context about our guidelines and best practices.
+You can copy and paste the following chatmode into your code editor to ensure it follows our guidelines or refer to it by linking it in your project specific custom chatmode.
 
 * [SN FED Chatmode](sn-fed.chatmode.md)
 
--- a/copilot/sn-fed.chatmode.md
+++ b/copilot/sn-fed.chatmode.md
@@ -8,7 +8,7 @@ You are an Expert Frontend Developer. Your guidance and code must strictly align
 ## Behavioral Rules
 
 ### Accessibility
-- Enforce [WCAG 2.1 AA](https://www.w3.org/TR/WCAG21/) accessibility standards for all UI and code recommendations.
+- Enforce [WCAG 2.2 AA](https://www.w3.org/TR/WCAG22/) accessibility standards for all UI and code recommendations.
 
 ---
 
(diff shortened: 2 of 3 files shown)


copilot/sn-fed.chatmode.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Description
+You are an Expert Frontend Developer. Your guidance and code must strictly align with the [Springer Nature Frontend Playbook](https://github.com/springernature/frontend-playbook). Always answer with best practices, code structure, and communication style as described in the playbook.
+
+---
+
+## Behavioral Rules
+
+### Accessibility
+- Enforce [WCAG 2.1 AA](https://www.w3.org/TR/WCAG21/) accessibility standards for all UI and code recommendations.
```
dotcode:
```suggestion
- Enforce [WCAG 2.2 AA](https://www.w3.org/TR/WCAG22/) accessibility standards for all UI and code recommendations.
```

dotcode (changes requested):
Requiring WCAG 2.2 AA, in oder to meet customer demand and get ahead of upcoming changes to legislation.

jshiell:
Let me also mention https://github.com/springernature/shared-prompts as a great place to share reusable prompts!

tulipdexter:
I was going to suggest that maybe frontend-tooling could be a better fit than the playbook.
But I was unware of the shiny shared-prompts repo ✨

Sreilys:
Thank you all for taking the time to look at this. @jshiell thanks for mentioning the shared prompts folder. From my understanding, chatmodes and promt files are slighly different. I think chatmode is a reusable chat persona/behavior for Copilot Chat (guardrails for the answer, scope, role) and Prompt provides a structured instructions (how to ask for something) for a specific request. Maybe we can have both chatmodes and prompts in that repo? Also now with opencode ai maybe also a shared folder for skills? I guess is a bit tricky to name those folders as every provider calls them in another way

jshiell:
> Maybe we can have both chatmodes and prompts in that repo? Also now with opencode ai maybe also a shared folder for skills? I guess is a bit tricky to name those folders as every provider calls them in another way

While it's called `shared-prompts`, the _goal_ is to gather shared resources for this tooling - whether than's prompts, agent definitions, skills, whatever. So I'd really encourage putting things in there, and we can panic about the structure if we get enough that it becomes an issue!

dotcode:
As shared-prompts is not a public repository, and these specific FED prompts have potentially great value if shared in frontend-playbook (all the accessibility stuff omg it will impact peoples lives), is the suggestion to remove this to shared-prompts, or rather to copy to shared-prompts?

… my preference would be to copy, obvs.

jshiell:
Copy is fine by me

copilot/README.md
```diff
@@ -0,0 +1,9 @@
+# Copilot
+
+If you must use AI, do so responsibly and in alignment with our frontend playbook.
+This chatmode give Copilot more context about our guidelines and best practices.
+You can copy and paste the following chatmode into your VSCode editor to ensure it follows our guidelines or refer to it by linking it in your project specific custom chatmode.
```
jon-stevens: Do we need to specify the editor?

copilot/sn-fed.chatmode.md
```diff
@@ -0,0 +1,160 @@
...
+
+**BEM Methodology**
+- Use Block, Element, Modifier (BEM) for CSS class naming:
+  - **Block:** `.c-block`
+  - **Element:** `.c-block__element`
+  - **Modifier:** `.c-block--modifier`
+- Never use elements inside elements (e.g., `.c-alert-box__close__button` is incorrect).
+- Always use the unmodified block alongside modifiers:  
+  `<div class="c-my-block c-my-block--modifier">`
+- Do not nest class selectors in preprocessors; keep selectors flat for searchability and clarity.
```
jon-stevens: I think 'Do not nest class selectors in preprocessed CSS' or just 'Do not nest class selectors' is clearer.

copilot/sn-fed.chatmode.md
```diff
@@ -0,0 +1,160 @@
...
+
+**How We Write CSS**
+- Each component should have its own CSS file.
+- Use variables for color, spacing, and font where possible.
+- Avoid using IDs for styling.
+- Use CSS custom properties for theming and maintainability.
+
```
jon-stevens, line 40: Is it worth adding something about use of utility classes? Maybe using utility classes from brand-context/elements where possible

copilot/README.md
Sreilys: No no, we don't have to, I will take that away and use a generic "code editor" term

copilot/sn-fed.chatmode.md
Sreilys: Amazing, will add. Thanks!

PR #503, opened 2026-01-28
Added recommendation to use punctuation.

cmuncharaz:
Edited the alt text guide to document the best practice of using punctuation, as discussed in the accessibility Slack channel.

### DIFF
--- a/accessibility/02-guide-to-alt-text.md
+++ b/accessibility/02-guide-to-alt-text.md
@@ -92,6 +92,7 @@ Now imagine "A cartoon representation of a yellow dog which has a very large hea
 * Don’t say ‘Image of…’ or ‘Picture of…’ etc. (It's nearly always wrong to say an "Image of", because all images are … images. It's extraneous information that the screenreader announces. However, it can help to know the type of image - i.e. is it a photo, or a screenshot, or a cartoon, depending on the surrounding content and context.)  
 * Don’t need to repeat information already given in the text on the page.  
 * Be as succinct as possible but include details as needed. 
+* Using punctuation in the text alternative makes the information easier to understand. Screen readers will pause at commas and periods. Add a full stop at the end.
 * Copyright info, image source, or other extra information should go in the caption, not the alt text.
 * Context is important e.g. it may be appropriate to note if an image is a photo or an artwork, whether a person is old or young, if the sky is stormy or clear, if that is relevant to the content. 
 * Think about how you would describe the image to a person who can’t see it.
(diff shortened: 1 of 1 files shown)


accessibility/02-guide-to-alt-text.md
```diff
@@ -92,6 +92,7 @@ Now imagine "A cartoon representation of a yellow dog which has a very large hea
 * Don’t say ‘Image of…’ or ‘Picture of…’ etc. (It's nearly always wrong to say an "Image of", because all images are … images. It's extraneous information that the screenreader announces. However, it can help to know the type of image - i.e. is it a photo, or a screenshot, or a cartoon, depending on the surrounding content and context.)  
 * Don’t need to repeat information already given in the text on the page.  
 * Be as succinct as possible but include details as needed. 
+* Using punctuation in the text alternative makes the information easier to understand. Screen readers will pause at commas and periods. Add a full stop at the end.
```
hollsk: I know this comes from the [WAI tips and tricks document](https://www.w3.org/WAI/tutorials/images/tips/), but I'd suggest more nuance here, as only some punctuation will help. I'd be tempted to rewrite more precisely to something like this:

```suggestion
* Using commas and periods in the text alternative makes the information easier to understand because screen readers will pause at them. Add a full stop at the end.
```

... and then add an extra bullet below that warns people that a lot of punctuation is silent on default settings (and that not every screen reader treats each punctuation mark the same way). Consider linking out to the [Deque punctuation and symbols post](https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/)? It's the most recent test suite I can think of (unless you know of a more contemporary one, which would be better).

accessibility/02-guide-to-alt-text.md
```diff
@@ -238,4 +239,4 @@ In the case of images of mathematical expressions, the alt text should describe
 #### Sufficient alt text
 
 * "The quantity of a plus b divided by d."
-* "Fraction where the numerator is a plus b and the denominator is d."
\ No newline at end of file
+* "Fraction where the numerator is a plus b and the denominator is d."
```
hollsk: how is "a plus b" read out? Does it sound like the letter "AY" being read out as the alphabet character, or does it sound like "AH", as in the indefinite article? 

Do the letters need to be capitalised to avoid the latter situation, if it happens?

hollsk (commented):
Thanks Carlos!

accessibility/02-guide-to-alt-text.md
cmuncharaz: We could link to "[How screen readers read special characters: an update](https://elevenways.be/en/articles/screenreaders-special-characters)", published by Eleven Ways. It's more contemporary (2023 VS 2014) and links to Deque's text.

accessibility/02-guide-to-alt-text.md
cmuncharaz: That's a really good point. There's a risk that some screen readers read out "a" as in the indefinite article. Capitalising the letters should help mitigating this risk, and we should probably make this clear in the guidance we provide.

PR #504, opened 2026-03-17
Caveat TLS 1.2 support section of browser support policy to address login and registration reliance on TLS 1.2

benjclark:
Thanks to @sky-jack for raising this. Perhaps worth us making this caveat in the Playbook?


sky-jack (approved):
All seems clear to me.

josebolos:
👋 Question: would this change not be covered already by the previous paragraph? As in, we're already preventing browsers that don't support TLS 1.2 from accessing the whole site(s) altogether, so I would expect that includes the authentication part too? Or is there a reason why authentication is mentioned separately?

sky-jack:
@josebolos  good point, Ben and I had earlier had a similar conversation. On reflection it is probably not necessary.

benjclark:
Yeah thanks @josebolos and @sky-jack . Jack and I were discussing and initially thought that raising awareness of the impact it has for user journeys behind authentication would be a useful thing. But in reality, as the first para covers, Core experience users on TLS 1.2 incompatible devices and browsers actually can't access any of the site anyway, so the distinction doesn't serve much purpose

